PR #20, opened 2018-07-26
Make Version a concrete model with GFK field to content

chaosk:


djangocms_versioning/cms_config.py
```diff
@@ -8,57 +8,36 @@
...
-                is_versioning_model = issubclass(model, BaseVersion)
-            except TypeError:
-                raise ImproperlyConfigured(
-                    "elements in versioning_models must be model classes")
-            if not is_versioning_model:
-                raise ImproperlyConfigured(
-                    "models in versioning_models must inherit from BaseVersion")
-        # If no exceptions raised, we can now add the versioning models
+                "versioning not defined as a list")
+        cms_config.versioning = VersionableList(cms_config.versioning)
```
monikasulik: Is this really necessary? I don't think it's good to be changing the class another app has provided on the fly like that. And I have some (very vague) scenarios in which this could go wrong. Atm versioning is a dependency of moderation and they actually do check our settings. This is because they need to check the app is not just enabled for moderation but also for versioning and they need to make sure that each content model that is registered for moderation has also been registered for versioning.
This means that if you change stuff in this class on the fly like that odd things could potentially happen (or even be dependent on what the order of versioning and moderation is in INSTALLED_APPS). It also means this is a breaking change for moderation ;) Although I don't think there's any way around that, but please be aware that they will probably murder us soon for breaking their code all the time ;) It would be a good idea to include them more in changes like this that could break stuff for them.

chaosk: True, changed to not do that.

djangocms_versioning/versionable.py
```diff
@@ -0,0 +1,65 @@
+from django.core.exceptions import ImproperlyConfigured
```
czpython: `datastructures.py`

djangocms_versioning/versionable.py
```diff
@@ -0,0 +1,65 @@
...
+        """
+        inner = self.content.objects.values(
+            self.grouper_field.name,
+        ).annotate(Max('pk')).values('pk__max')
+        return self.content.objects.filter(
+            id__in=inner,
+        )
+
+
+class VersionableList(list):
```
czpython: no need for this

djangocms_versioning/versionable.py
```diff
@@ -0,0 +1,65 @@
+from django.core.exceptions import ImproperlyConfigured
+from django.db.models import Max
+
+
+class Versionable:
```
czpython: item or model?

djangocms_versioning/versionable.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def __init__(self, grouper, content, grouper_field_name=None):
```
czpython: don't make grouper_field_name optional

djangocms_versioning/versionable.py
```diff
@@ -0,0 +1,65 @@
...
+    def distinct_groupers(self):
+        """Returns a queryset of `self.content` objects with unique
+        grouper objects.
+
+        Useful for listing, e.g. all Polls.
+        """
+        inner = self.content.objects.values(
+            self.grouper_field.name,
+        ).annotate(Max('pk')).values('pk__max')
+        return self.content.objects.filter(
```
czpython: one line

djangocms_versioning/models.py
```diff
@@ -1,51 +1,44 @@
...
-
 
-class BaseVersion(models.Model):
+class Version(models.Model):
     # Following fields are always copied from original Version
     COPIED_FIELDS = ['label']
 
     label = models.TextField()
     created = models.DateTimeField(auto_now_add=True)
+    content_type = models.ForeignKey(
```
czpython: can't be optional

tests/test_admin.py
```diff
@@ -124,7 +119,9 @@ def test_poll_version_is_added_for_change_false(self):
             pc1 = factories.PollContentFactory()
             request = RequestFactory().get('/admin/polls/pollcontent/')
             model_admin.save_model(request, pc1, form=None, change=False)
-            check_obj = PollVersion.objects.get(content=pc1)
+            check_obj = Version.objects.get(
+                content_type=ContentType.objects.get_for_model(pc1),
+                object_id=pc1.pk)
```
czpython: match with variable start

djangocms_versioning/admin.py
```diff
@@ -1,9 +1,17 @@
...
+from django.contrib.admin.views.main import ChangeList
+from django.contrib.contenttypes.models import ContentType
 from django.urls import reverse
 from django.utils.html import format_html
 from django.utils.translation import ugettext_lazy as _
 
+from .models import Version
+
+
+GROUPER_PARAM = 'grouper'
```
monikasulik: Are we assuming that the grouper fk on the content model will always be called "grouper"? Also if it's a constant then it probably needs to be in constants.py, right? :)

djangocms_versioning/admin.py
```diff
@@ -16,23 +24,45 @@ def save_model(self, request, obj, form, change):
...
-            queryset.model]
-        filter_name = '{}__in'.format(version_model.__name__.lower())
-        latest_versions = version_model.objects.distinct_groupers()
-        return queryset.filter(**{filter_name: latest_versions})
+        versioning_extension = apps.get_app_config('djangocms_versioning').cms_extension
+        versionable = versioning_extension.versionables_by_content[queryset.model]
+        return queryset.filter(pk__in=versionable.distinct_groupers())
 
 
+class VersionChangeList(ChangeList):
```
monikasulik: Maybe not a discussion for this PR, but I think this is becoming quite complex and am wondering if it wouldn't be simpler for maintainability to create a custom view for the version list rather than modify the django changelist for this purpose. In my admin experiences so far, it's been the case that as an application grows, a custom view turns out to be much easier to maintain.

djangocms_versioning/models.py
```diff
@@ -1,51 +1,42 @@
...
-class BaseVersionQuerySet(models.QuerySet):
+class VersionQuerySet(models.QuerySet):
 
     def for_grouper(self, grouper, extra_filters=None):
         """Returns all `Version`s for specified grouper object.
 
         Additional filters can be passed via extra_filters, for example
         if version model has a FK to content object that is translatable,
-        passing `Q(content__language='en')` will return only versions
+        passing `Q(language='en')` will return only versions
```
monikasulik: How come just language would work now? We haven't changed the version model to include language, right?

tests/test_admin.py
```diff
@@ -134,24 +132,29 @@ def test_poll_version_is_not_added_for_change_true(self):
...
-        check_obj_exist = PollVersion.objects.filter(content=pc2).exists()
+        check_obj_exist = Version.objects.filter(
+            content_type=ContentType.objects.get_for_model(pc2),
+            object_id=pc2.pk).exists()
         self.assertFalse(check_obj_exist)
 
     def test_blogpost_version_is_added_for_change_false(self):
         model_admin = self._get_admin_class_obj(BlogContent)
-        bc1 = factories.BlogContentFactory()
+        pc2 = factories.BlogContentFactory()
```
monikasulik: Why is this getting changed here? bc would stand for blog content, but what would pc stand for in this case?

djangocms_versioning/admin.py
chaosk: This is just a query param name, it's not related to the name of grouper FK field on content model. Might move it to `constants.py`, as it will be used to build "Manage versions" URL.

djangocms_versioning/admin.py
chaosk: @czpython ?

chaosk: @monikasulik Talked to Paulo, custom view would require maintaining admin UI, so that breadcrumbs would still have to work, filtering and sorting would have to be reimplemented. So for now, we're sticking with regular changelist view

czpython:
🥇

tests/test_admin.py
```diff
@@ -219,53 +222,29 @@ def test_records_filtering_is_generic(self):
...
-            'content',
-            self.site._registry[PollVersion].list_select_related,
-        )
-
     def test_content_link(self):
-        version = factories.PollContentWithVersionFactory(text='test4').pollversion
+        pc = factories.PollContentWithVersionFactory(text='test4')
+        version = self.model.objects.get(
+            content_type=ContentType.objects.get_for_model(pc),
+            object_id=pc.pk)
```
monikasulik: Not particularly important, but I think atm you could just do `version = factories.PollVersionFactory(content__text='test4')` and then refer to content as `version.content`. It would save some lines and be a little easier to read probably.

tests/test_admin.py
```diff
@@ -219,53 +222,29 @@ def test_records_filtering_is_generic(self):
...
         self.assertEqual(
-            self.site._registry[PollVersion].content_link(version),
+            self.site._registry[self.model].content_link(version),
```
monikasulik: I'm not sure we really need `self.model`. To me at least it would be clearer to just have `Version` here. Otherwise to understand the test I have to check `setUp` to see what `self.model` is.

tests/test_datastructures.py
```diff
@@ -0,0 +1,37 @@
...
+    PollVersionFactory,
+)
+from djangocms_versioning.test_utils.polls.models import PollContent
+
+
+class VersionableItemTestCase(CMSTestCase):
+
+    def setUp(self):
+        self.initial_version = PollVersionFactory()
+        AnswerFactory.create_batch(
```
monikasulik: Do you really need answers in these tests? I think they're needed for the copy tests, but not sure they add anything here?

djangocms_versioning/models.py
```diff
@@ -116,6 +77,9 @@ def copy(self):
         m2m_cache = {}
 
         for field in relation_fields:
+            # FIXME .copy() is broken after GFK changes
```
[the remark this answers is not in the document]
chaosk: Was using `.copy()` in `test_datastructures.py`, removed that.

djangocms_versioning/admin.py
```diff
@@ -16,23 +24,49 @@ def save_model(self, request, obj, form, change):
...
-            queryset.model]
-        filter_name = '{}__in'.format(version_model.__name__.lower())
-        latest_versions = version_model.objects.distinct_groupers()
-        return queryset.filter(**{filter_name: latest_versions})
+        versioning_extension = apps.get_app_config('djangocms_versioning').cms_extension
+        versionable = versioning_extension.versionables_by_content[queryset.model]
+        return queryset.filter(pk__in=versionable.distinct_groupers())
+
+
+class VersionChangeList(ChangeList):
```
monikasulik: Correct me if I'm wrong, but I don't think this is tested at all. Either add an integration test and/or unit tests for each method. My gut feeling it that the best way to go is integration tests for the main functionality (i.e. adding grouper to the url query or not adding it and if we get the results we're expecting - I think the django client gives a way to access the queryset). And then unit tests for `IncorrectLookupParameters`.

djangocms_versioning/admin.py
```diff
@@ -46,7 +80,12 @@ class VersionAdmin(admin.ModelAdmin):
         'label',
     )
     list_display_links = None
-    list_select_related = ('content',)
+
+    def get_queryset(self, request):
```
monikasulik: If you do the integration test mentioned above then you'll have this bit covered as well I think (and `get_changelist`). But if you don't do the above as an integration test then you should make sure these methods are tested too.

tests/test_datastructures.py
```diff
@@ -0,0 +1,37 @@
...
+            2, poll_content=self.initial_version.content)
+
+    def test_distinct_groupers(self):
+        self.initial_version.copy()
+        poll2_version = PollVersionFactory()
+        poll2_version.copy()
+        poll2_version.copy()
+
```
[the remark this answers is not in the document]
chaosk: Agreed!

PR #21, opened 2018-07-27
State management

monikasulik:
This can't be merged until the GFK stuff is done. To see a more friendly diff in the meantime, see the Fidelity PR: https://github.com/FidelityInternational/djangocms-versioning/pull/10 (this shows the changes against the gfk branch)


djangocms_versioning/test_utils/blogpost/cms_config.py
```diff
@@ -1,8 +1,19 @@
...
-    versioning_models = [BlogPostVersion, CommentVersion]
+    versioning = [
+        Versionable(
+            grouper=BlogPost,
+            content=BlogContent,
+        ),
+        Versionable(
+            grouper=BlogPost,
+            content=Comment,
+        )
```
chaosk: I removed versioning definition for comment, because I'm not sure it's correct - having a BlogPost as a grouper for Comment doesn't make Comment model versionable - it doesn't allow for a single comment under a post to have multiple versions.

djangocms_versioning/models.py
```diff
@@ -1,51 +1,74 @@
...
+        null=True,
+        on_delete=models.SET_NULL,
+    )
+    object_id = models.PositiveIntegerField(blank=True, null=True)
+    content = GenericForeignKey('content_type', 'object_id')
+
+    objects = VersionQuerySet.as_manager()
+
+    # States
+    ARCHIVED = 'Archived'
```
chaosk: Possibly move those to `constants.py` module (and rename to VERSION_STATE_ARCHIVED/STATE_ARCHIVED/VERSION_ARCHIVED). I also prefer string choice values to be lowercase, making it easier to manually filter by specified value while debugging, what do you think of that?

tests/test_states.py
```diff
@@ -0,0 +1,68 @@
...
+from cms.test_utils.testcases import CMSTestCase
+
+from djangocms_versioning.test_utils import factories
+from djangocms_versioning.models import Version
+
+
+class TestVersionState(CMSTestCase):
+
+    def test_direct_modification_of_state_not_allowed(self):
+        version = factories.PollVersionFactory(state='Draft')
```
chaosk: Use state constants that are currently defined in the model

PR #23, opened 2018-07-30
State change tracking

monikasulik:
This contains changes from #20 and #21 so can't be merged until those two are merged. In the meantime for a friendlier diff see https://github.com/FidelityInternational/djangocms-versioning/pull/11


djangocms_versioning/test_utils/factories.py
```diff
@@ -1,8 +1,23 @@
 import factory
 from factory.fuzzy import FuzzyText
 
-from .blogpost.models import BlogContent, BlogPost, BlogPostVersion
-from .polls.models import Answer, Poll, PollContent, PollVersion
+from django.contrib.auth.models import User
```
chaosk: ``
from django.contrib.auth import get_user_model
...

    class Meta:
        model = get_user_model()
```

monikasulik: I can do this change if you like, but I would argue it's better as it is. It's usually better in test setup/factories if we're as explicit as possible.
Also, because these are unit tests for versioning in isolation, AFAIK it is highly unlikely this will ever be something other than `User` (which is the main reason I don't particularly care if we change it to `get_user_model` - if there was a chance that the user model would change then I would argue it's better if somebody makes this change in tests explicitly to check that's definitely correct for all tests).

PR #24, opened 2018-08-01
WIP: Adjust copy method to work with new version model

monikasulik:


monikasulik:
This task has now been separated into three. Two related PRs are #33 and #34

PR #25, opened 2018-08-02
Versioning integration docs

chaosk:


README.rst
```diff
@@ -39,20 +39,22 @@ There are four key terms or classes you need to understand in order to implement
...
 
 **Content** - a single versioned object, must have a FK to a *Grouper* object - mandatory. If you are implementing Versioning on an existing content model then this will be that model. All you need to add is the foreign key to your *Grouper* model.
 
-**ContentExtension** - an object which tracks additional information / relationships associated with the content - optional. This can be any model associated with your *Content* model. It may be a 121 or 12M or M2M relationship. It's your responsibility to implement versioning for these extensions via the Version::copy_content method in your implementation of the *Version* class described above.
+**ContentExtension** - an object which tracks additional information / relationships associated with the content - optional. This can be any model associated with your *Content* model. It may be a 121 or 12M or M2M relationship. It's your responsibility to implement versioning for these extensions by providing copy functions in your VersionableItem definitions.
 
 The key things needed in order to implement DjangoCMS Versioning are:
```
monikasulik: I, personally, find this quite difficult to read and I find it a bit repetitive as I think you're basically explaining this more clearly with code examples lower down. For me, personally, code examples are always preferable to long blocks of text. So I think we could remove this section completely. What do you think?

docs/versioning_integration.md
```diff
@@ -0,0 +1,60 @@
...
+Versioning needs to be aware of these models. This can be done in `cms_config.py` file:
+
+```python
+# polls/cms_config.py
+class PollsCMSConfig(CMSAppConfig):
+    djangocms_versioning_enabled = True  # -- 1
+    versioning = [
+        VersionableItem(   # -- 2
+            content_model=PollContent,
+            grouper_field_name='poll',
```
monikasulik: Please make sure I update this when you code review my copy method stuff :D It's the sort of thing I'm likely to forget ;)

README.rst
```diff
@@ -39,20 +39,22 @@ There are four key terms or classes you need to understand in order to implement
...
 
 **Content** - a single versioned object, must have a FK to a *Grouper* object - mandatory. If you are implementing Versioning on an existing content model then this will be that model. All you need to add is the foreign key to your *Grouper* model.
 
-**ContentExtension** - an object which tracks additional information / relationships associated with the content - optional. This can be any model associated with your *Content* model. It may be a 121 or 12M or M2M relationship. It's your responsibility to implement versioning for these extensions via the Version::copy_content method in your implementation of the *Version* class described above.
+**ContentExtension** - an object which tracks additional information / relationships associated with the content - optional. This can be any model associated with your *Content* model. It may be a 121 or 12M or M2M relationship. It's your responsibility to implement versioning for these extensions by providing copy functions in your VersionableItem definitions.
```
monikasulik: AFAIK, the only place in the codebase where we refer to `ContentExtension` is here. We do need to tell them about the need to provide copy functions, but I think it's better to do that in the file below (again, remind me of that when I get the copy PR to you!) with code examples. This term, to me, is rather confusing.

docs/versioning_integration.md
chaosk: Will do!

docs/versioning_integration.md
```diff
@@ -0,0 +1,60 @@
...
+
+
+class PostContent(models.Model):
+    post = models.ForeignKey(Post, on_delete=models.CASCADE)
+    title = models.CharField(max_length=100)
+    text = models.TextField()
+```
+
+This structure is not only compatible with versioning,
+but also works when versioning is not enabled.
```
damilare: Is there supposed to be an explanation for when versioning is not enabled for example, what happens to ```Post``` and ```PostContent``` ?

docs/versioning_integration.md
```diff
@@ -0,0 +1,60 @@
...
+    djangocms_versioning_enabled = True  # -- 1
+    versioning = [
+        VersionableItem(   # -- 2
+            content_model=PollContent,
+            grouper_field_name='poll',
+        ),
+    ]
+```
+
+1. This line instructs CMS to pass this configuration to Versioning
```
mmoravcik: Don't think it is clear to tell if this line means that versioning is enabled for this app or that the app has versioning support? Or both?

docs/versioning_integration.md
```diff
@@ -0,0 +1,60 @@
...
+`Post` becomes a **grouper** model and `PostContent` becomes a **content** model.
+
+Versioning needs to be aware of these models. This can be done in `cms_config.py` file:
+
+```python
+# polls/cms_config.py
+class PollsCMSConfig(CMSAppConfig):
+    djangocms_versioning_enabled = True  # -- 1
+    versioning = [
+        VersionableItem(   # -- 2
```
mmoravcik: It would be good to include where to import/get this from as an example?

docs/versioning_integration.md
chaosk: This line means that app supports versioning and that if versioning is installed, CMS will pass this configuration to versioning. I guess we could link to CMS docs regarding new app registration here.

docs/versioning_integration.md
chaosk: Addressed

docs/versioning_integration.md
chaosk: Addressed

docs/versioning_integration.md
```diff
@@ -0,0 +1,79 @@
...
+            content_model=PostContent,
+            grouper_field_name='post',
+        ),
+    ]
+```
+
+1. This line instructs CMS to pass this configuration to Versioning
```
monikasulik: How about we change this to `This must be set to True for versioning to read your cms config` - that might be clearer?

PR #28, opened 2018-08-03
Add is_content_model_versioned to VersioningCMSExtension

chaosk:


djangocms_versioning/cms_config.py
```diff
@@ -18,6 +18,11 @@ def __init__(self):
     def versionables_by_content(self):
         return {versionable.content_model: versionable for versionable in self.versionables}
 
+    def is_content_model_versioned(self, content_model):
```
monikasulik: Make sure moderation is aware of this as it will be useful to them.

tests/test_cms_config.py
```diff
@@ -90,6 +90,28 @@ def test_handle_content_admin_classes(self):
...
 
+    def test_is_content_model_versioned(self):
+        """Test that is_content_model_versioned returns True for
+        content model that's versioned
+        """
+        extensions = VersioningCMSExtension()
+        cms_config = Mock(
+            spec=[], djangocms_versioning_enabled=True,
+            versioning=[VersionableItem(content_model=PollContent, grouper_field_name='poll')])
+        extensions.handle_versioning_setting(cms_config)
```
monikasulik: As below, I suggest you not use `handle_versioning_setting` for data set up. You can set it up like this instead:
```
extensions = VersioningCMSExtension()
extensions.versionables = [VersionableItem(content_model=PollContent, grouper_field_name='poll')]
```

PR #29, opened 2018-08-06
Create separate version admin classes dynamically for every registered content model

chaosk:


djangocms_versioning/cms_config.py
```diff
@@ -43,6 +43,17 @@ def handle_admin_classes(self, cms_config):
             [versionable.content_model for versionable in cms_config.versioning],
         )
 
+    def handle_version_admin(self, cms_config):
+        """Registers version admin for all registered content types
+        with default filtering by content type applied.
+        """
+        for versionable in cms_config.versioning:
+            register_versionadmin_proxy(
```
czpython: pass versionable item here

djangocms_versioning/datastructures.py
```diff
@@ -11,6 +14,25 @@ def __init__(self, content_model, grouper_field_name):
...
+    def version_proxy(self):
+        class Meta:
+            proxy = True
+        ProxyVersion = type(
+            self.grouper_field.remote_field.model.__name__ + 'Version',
+            (Version, ),
+            {
+                'Meta': Meta,
+                '__module__': __name__,
+                '_content_model': self.content_model,
```
czpython: `_source_model`

djangocms_versioning/datastructures.py
```diff
@@ -11,6 +14,25 @@ def __init__(self, content_model, grouper_field_name):
     def _get_grouper_field(self):
         return self.content_model._meta.get_field(self.grouper_field_name)
 
+    @cached_property
+    def version_proxy(self):
```
monikasulik: I would change this to `version_model_proxy` as it's actually very important to the understanding of this code that it returns a model class. And it would be good to add a docstring as well as it's really not a straightforward property! :)

djangocms_versioning/forms.py
```diff
@@ -0,0 +1,20 @@
...
+from django import forms
+from django.apps import apps
+
+
+@lru_cache()
+def grouper_form_factory(content_model):
+    versioning_extension = apps.get_app_config('djangocms_versioning').cms_extension
+    versionable = versioning_extension.versionables_by_content[content_model]
+    return type(
+        versionable.grouper_field.remote_field.model.__name__ + 'GrouperForm',
```
monikasulik: Can be shortened to `versionable.grouper_model.__name__` I think

djangocms_versioning/forms.py
```diff
@@ -0,0 +1,20 @@
+from functools import lru_cache
+
+from django import forms
+from django.apps import apps
+
+
+@lru_cache()
+def grouper_form_factory(content_model):
```
monikasulik: A docstring would help immensely for readability.

djangocms_versioning/helpers.py
```diff
@@ -43,3 +44,28 @@ def replace_admin_for_models(models, admin_site=None):
...
+        return
+
+    class ProxiedAdmin(VersionAdmin):
+
+        def get_queryset(self, request):
+            content_type = ContentType.objects.get_for_model(self.model._content_model)
+            return super().get_queryset(request).filter(
+                content_type=content_type,
+            )
+    ProxiedAdmin.__name__ = grouper_name + VersionAdmin.__name__
```
monikasulik: Isn't `VersionAdmin.__name__` the same as `'VersionAdmin'` in this case? If yes then IMO `'VersionAdmin'` would be more readable.

djangocms_versioning/helpers.py
```diff
@@ -43,3 +44,28 @@ def replace_admin_for_models(models, admin_site=None):
         except KeyError:
             continue
         _replace_admin_for_model(modeladmin, admin_site)
+
+
+def register_versionadmin_proxy(version_proxy, grouper_name, admin_site=None):
+    """
```
monikasulik: Include some info in the docstring about what this function actually does as well.

tests/test_admin.py
```diff
@@ -229,9 +231,25 @@ class AdminRegisterVersionTestCase(CMSTestCase):
     def test_register_version_admin(self):
         """Test that VersionAdmin is registered for Version model
```
monikasulik: I think the docstring needs to be updated.

tests/test_admin.py
```diff
@@ -229,9 +231,25 @@ class AdminRegisterVersionTestCase(CMSTestCase):
...
+        doesn't do anything.
+        """
+        existing_admin = type('TestAdmin', (admin.ModelAdmin, ), {})
+        site = admin.AdminSite()
+        site.register(PollContent, existing_admin)
+
+        with patch.object(site, 'register') as mock:
+            register_versionadmin_proxy(PollContent, 'foo', site)
+
+        mock.assert_not_called()
```
monikasulik: If a warning is added then this should probably also test that the warning is being logged.

djangocms_versioning/admin.py
```diff
@@ -53,27 +53,17 @@ def get_queryset(self, request):
...
             grouper = int(request.GET.get(GROUPER_PARAM))
         except (TypeError, ValueError):
-            raise IncorrectLookupParameters("Invalid grouper")
-        content_type_id = request.GET.get('content_type_id')
-        try:
-            content_type = ContentType.objects.get(pk=content_type_id)
-            model = content_type.model_class()
-        except (ContentType.DoesNotExist, ValueError):
-            raise IncorrectLookupParameters("Invalid content_type_id")
+            return qs
```
monikasulik: Correct me if I'm wrong, but I don't think you have a test for this. I.e. you check that in this case a grouper form is displayed, but you never check that a full queryset is returned.

djangocms_versioning/cms_config.py
```diff
@@ -43,6 +43,17 @@ def handle_admin_classes(self, cms_config):
             [versionable.content_model for versionable in cms_config.versioning],
         )
 
+    def handle_version_admin(self, cms_config):
```
monikasulik: I don't think you have any cms config tests for this. We should really be checking that when a versionable is defined in cms_config.py it is appropriately registered in the admin.

djangocms_versioning/datastructures.py
monikasulik: I don't think this property has been tested. It's a complex enough property that I do think it should have a test.

djangocms_versioning/datastructures.py
```diff
@@ -11,6 +14,25 @@ def __init__(self, content_model, grouper_field_name):
...
+            {
+                'Meta': Meta,
+                '__module__': __name__,
+                '_content_model': self.content_model,
+            },
+        )
+        return ProxyVersion
+
+    @property
+    def grouper_model(self):
```
monikasulik: Not as important as this property is simpler, but I think it would be good to have a test for this as well.

djangocms_versioning/forms.py
monikasulik: This doesn't seem to have a unit test.

djangocms_versioning/helpers.py
monikasulik: You're not really testing this line (the setting of the name) anywhere I think. I'd add an assert to one of your unit tests.

djangocms_versioning/forms.py
```diff
@@ -0,0 +1,20 @@
...
+    versioning_extension = apps.get_app_config('djangocms_versioning').cms_extension
```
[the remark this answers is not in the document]
chaosk: @czpython?

djangocms_versioning/helpers.py
chaosk: This way makes it easier to rename the class, if there's ever a need for that.

djangocms_versioning/helpers.py
```diff
@@ -43,3 +44,35 @@ def replace_admin_for_models(models, admin_site=None):
...
+
+    :param versionable: VersionableItem instance
+    :param admin_site: AdminSite instance
+    """
+    if admin_site is None:
+        admin_site = admin.site
+
+    if versionable.version_model_proxy in admin_site._registry:
+        # Attempting to register the proxy again is a no-op.
+        return
```
monikasulik: There's still no warning logged here (or exception thrown - depending on how we think it's better to handle). I do think doing this silently isn't too great.

test_settings.py
```diff
@@ -3,7 +3,8 @@
     'INSTALLED_APPS': [
         'djangocms_versioning',
         'djangocms_versioning.test_utils.polls',
-        'djangocms_versioning.test_utils.blogpost'
+        'djangocms_versioning.test_utils.blogpost',
+        'djangocms_versioning.test_utils.non_configured',
```
monikasulik: I would just call this app people and possibly not put it in INSTALLED_APPS at all? I think the relationships app I created does create the models even without putting it in INSTALLED_APPS.

tests/test_admin.py
```diff
@@ -240,6 +258,8 @@ def setUp(self):
         self.site = admin.AdminSite()
         self.site.register(Version, VersionAdmin)
         self.superuser = self.get_superuser()
+        versioning_extension = apps.get_app_config('djangocms_versioning').cms_extension
+        self.versionable = versioning_extension.versionables_by_content[PollContent]
```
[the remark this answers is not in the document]
monikasulik: This hasn't been acted on yet I think.

tests/test_admin.py
```diff
@@ -268,64 +288,58 @@ def test_content_link(self):
...
+            response = self.client.get(self.get_admin_url(self.versionable.version_proxy, 'delete', 1))
         self.assertEqual(response.status_code, 403)
 
 
 class VersionChangeListTestCase(CMSTestCase):
 
     def setUp(self):
         self.superuser = self.get_superuser()
+        versioning_extension = apps.get_app_config('djangocms_versioning').cms_extension
+        self.versionable = versioning_extension.versionables_by_content[PollContent]
```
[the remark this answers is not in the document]
monikasulik: Same here

djangocms_versioning/admin.py
```diff
@@ -108,6 +100,36 @@ def content_link(self, obj):
...
+                opts.model_name,
+            )))
+        return super().changelist_view(request, extra_context)
+
+    def get_urls(self):
+        info = self.model._meta.app_label, self.model._meta.model_name
+        return [
+            url(
+                r'^grouper/$',
+                self.admin_site.admin_view(self.grouper_form_view),
```
monikasulik: You probably should be testing for the admin permissions on this url (it's the sort of thing that's easy to miss).

tests/test_datastructures.py
```diff
@@ -43,3 +49,41 @@ def test_for_grouper(self):
...
             ordered=False
         )
+
+    def test_grouper_model(self):
+        versionable = VersionableItem(content_model=PollContent, grouper_field_name='poll')
+
+        self.assertEqual(versionable.grouper_model, Poll)
+
+
+class VersionableItemProxyModelTestCase(CMSTestCase):
```
monikasulik: Now that you have all these tearDown methods, do you still need the person app?

chaosk: Yep, otherwise we still have the problem with creating proxies that have already been registered during project startup.

tests/test_forms.py
```diff
@@ -0,0 +1,33 @@
...
+class GrouperFormTestCase(CMSTestCase):
+
+    def tearDown(self):
+        grouper_form_factory.cache_clear()
+
+    def test_factory(self):
+        pv = factories.PollVersionFactory()
+
+        form_class = grouper_form_factory(PollContent)
+        form = form_class()
```
monikasulik: I don't think you need to do this. You'll just need to check the form fields against `form.base_fields` instead of `form.fields`

tests/test_admin.py
```diff
@@ -296,6 +294,14 @@ def test_content_link(self):
             ),
         )
 
+
+class VersionAdminViewTestCase(CMSTestCase):
+
+    def setUp(self):
+        self.superuser = self.get_superuser()
+        versioning_extension = apps.get_app_config('djangocms_versioning').cms_extension
+        self.versionable = versioning_extension.versionables_by_content[PollContent]
```
monikasulik: As previously mentioned, I think it's better to just import the versionable from polls/cms_config.py rather than rely on this to set up data.

monikasulik: Here's an example: https://github.com/divio/djangocms-versioning/blob/master/tests/test_cms_config.py#L128 This avoids set up having a dependency on our code and only depends on test_utils which is fine.

tests/test_forms.py
```diff
@@ -0,0 +1,33 @@
...
+        self.assertIn(forms.Form, form_class.mro())
+        self.assertEqual(form_class.__name__, 'PollContentGrouperForm')
+        self.assertIn('grouper', form.fields)
+        self.assertIn(
+            (pv.content.poll.pk, str(pv.content.poll)),
+            form.fields['grouper'].choices,
+        )
+
+    def test_factory_cache(self):
+        self.assertEqual(
```
monikasulik: I think it's worth adding a comment here to explain that if it wasn't cached id would return different values.

djangocms_versioning/admin.py
```diff
@@ -99,18 +101,35 @@ def content_link(self, obj):
...
             opts=self.model._meta,
-            form=grouper_form_factory(self.model._content_model)(),
+            form=grouper_form_factory(self.model._source_model)(),
         )
         return render(request, 'djangocms_versioning/admin/grouper_form.html', context)
 
     def changelist_view(self, request, extra_context=None):
         if not request.GET:
-            return self.grouper_form_view(request)
+            opts = self.model._meta
```
monikasulik: A comment on what's happening here would help readability. Something like "We expect to have a grouper id in GET params. If we don't then redirect to make the user choose a grouper"

djangocms_versioning/forms.py
monikasulik: :+1:

djangocms_versioning/datastructures.py
```diff
@@ -11,6 +14,32 @@ def __init__(self, content_model, grouper_field_name):
...
+
+        model_name = self.content_model.__name__ + 'Version'
+
+        class Meta:
+            proxy = True
+        ProxyVersion = type(
+            model_name,
+            (Version, ),
+            {
+                'Meta': Meta,
```
czpython: use type

djangocms_versioning/datastructures.py
```diff
@@ -11,6 +14,32 @@ def __init__(self, content_model, grouper_field_name):
     def _get_grouper_field(self):
         return self.content_model._meta.get_field(self.grouper_field_name)
 
+    @cached_property
+    def version_model_proxy(self):
+        """Returns a dynamically created proxy model class to Version.
+        It's used for creating separate version model classes for each
+        content type.
+        """
+
```
czpython: remove newline

djangocms_versioning/helpers.py
```diff
@@ -43,3 +46,41 @@ def replace_admin_for_models(models, admin_site=None):
...
+
+    :param versionable: VersionableItem instance
+    :param admin_site: AdminSite instance
+    """
+    if admin_site is None:
+        admin_site = admin.site
+
+    if versionable.version_model_proxy in admin_site._registry:
+        # Attempting to register the proxy again is a no-op.
+        warnings.warn(
```
monikasulik: Not very important but I would usually also add an assert somewhere where I'd mock the warn method and check it's being called with the message I expect.

monikasulik: :+1:

PR #35, opened 2018-08-09
Allow custom copy methods

monikasulik:


djangocms_versioning/models.py
```diff
@@ -1,3 +1,4 @@
+from django.apps import apps
```
chaosk: Add unique_together constraint on (content_type, object_id)

djangocms_versioning/datastructures.py
```diff
@@ -6,10 +6,11 @@
```
chaosk: Update docs to reflect new changes.

monikasulik: Done

djangocms_versioning/models.py
monikasulik: Done

czpython:
This is great! Perfect execution! 💯 🥇 💯

tests/test_models.py
```diff
@@ -44,3 +73,21 @@ def test_content_object_gets_duplicated(self):
             original_version.content.poll,
             new_version.content.poll,
         )
+
+    def test_the_copy_method_is_configurable(self):
+        """When copying, the new version object should have a new
```
damilare: docs in tests, suger!

damilare:
All Hail Lady Monika, you have done incredible well!!! Great Stuff!! :rainbow: :unicorn: :dragon: :dragon_face: :performing_arts:

djangocms_versioning/datastructures.py
```diff
@@ -6,10 +6,11 @@
 
 class VersionableItem:
 
-    def __init__(self, content_model, grouper_field_name):
+    def __init__(self, content_model, grouper_field_name, copy_function):
```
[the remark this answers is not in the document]
monikasulik: We actually had a discussion about this with Krzysztof. The reason is that this way the developer who puts versioning into their project is more conscious of there being such a thing as a copy function. I'm anticipating that they will need to change it perhaps even 80% of the time or else have strange things happen, so I wanted to make this as visible as possible.

PR #36, opened 2018-08-10
Change to Archived state endpoint

monikasulik:


djangocms_versioning/admin.py
```diff
@@ -99,6 +104,19 @@ def content_link(self, obj):
     content_link.short_description = _('Content')
     content_link.admin_order_field = 'content'
 
+    def state_actions(self, obj):
+        """Display links to state change endpoints
+        """
+        archive_url = reverse('admin:djangocms_versioning_{model}_archive'.format(
+            model=self.model.__name__.lower(),
```
chaosk: Use `obj._meta.app_label` and `obj._meta.model_name`

djangocms_versioning/admin.py
```diff
@@ -110,6 +128,35 @@ def grouper_form_view(self, request):
...
+        # FIXME: We should be using POST only for this, but some frontend
+        # issues need to be solved first. The code below just needs to
+        # be uncommented and a test is also already written (but currently
+        # being skipped) to handle the POST-only approach
+
+        # This view always changes data so only POST requests should work
+        # if request.method != 'POST':
+        #     raise Http404
+
+        version = self.model.objects.get(pk=pk)
```
chaosk: https://github.com/django/django/blob/master/django/contrib/admin/options.py#L1845-L1849

djangocms_versioning/models.py
```diff
@@ -24,7 +25,37 @@ class Version(models.Model):
...
+        if self.state == constants.DRAFT:
+            pks_for_grouper = self.versionable.for_grouper(
+                self.grouper).values_list('pk', flat=True)
+            to_archive = Version.objects.exclude(pk=self.pk).filter(
+                state=constants.DRAFT, object_id__in=pks_for_grouper)
+            for version in to_archive:
+                version.archive(self.created_by)
+                version.save()
+
+    @property
```
chaosk: Perhaps test these two properties

djangocms_versioning/admin.py
monikasulik: `obj._meta.model_name` doesn't actually work because it seems to return the non-proxy model name

djangocms_versioning/admin.py
monikasulik: I'm not really following what you want me to do or what that code does? There's no comments there as to what TO_FIELD_VAR actually means or does and didn't find docs on it either. And not sure in what context fields would be disallowed?

PR #53, opened 2018-08-29
Versioning PageContent

monikasulik:


djangocms_versioning/cms_config.py
```diff
@@ -34,10 +36,16 @@ def handle_versioning_setting(self, cms_config):
         if not isinstance(cms_config.versioning, collections.abc.Iterable):
             raise ImproperlyConfigured(
                 "versioning not defined as an iterable")
+        registered_models = []
         for versionable in cms_config.versioning:
             if not isinstance(versionable, VersionableItem):
                 raise ImproperlyConfigured(
                     "{!r} is not a subclass of djangocms_versioning.datastructures.VersionableItem".format(versionable))
+            if versionable.content_model in registered_models:
```
chaosk: Doesn't this only check if a content model is registered twice in a single cms config?

PR #55, opened 2018-08-30
Inject published-only manager and add version renderer

chaosk:


djangocms_versioning/test_utils/polls/models.py
```diff
@@ -27,3 +29,10 @@ class Answer(models.Model):
 
     def __str__(self):
         return self.text
+
+
+class PollPlugin(CMSPlugin):
+    poll = models.ForeignKey(Poll, on_delete=models.CASCADE)
+
+    def __str__(self):
+        return str(self.poll) or str(self.pk)
```
czpython: bit strange. why the fallback?

djangocms_versioning/helpers.py
```diff
@@ -84,3 +88,49 @@ def get_queryset(self, request):
...
+    ]
+    model.add_to_class('objects', manager)
+
+
+def inject_generic_relation_to_version(model):
+    model.add_to_class('versions', GenericRelation(Version))
+
+
+@contextmanager
+def override_default_manager(model, manager):
```
czpython: what does the manager param do here? its overridden by the forloop below

djangocms_versioning/plugin_rendering.py
```diff
@@ -0,0 +1,54 @@
...
+from .constants import DRAFT, PUBLISHED
+
+
+def get_versionable_for_grouper(model):
+    versioning_extension = apps.get_app_config('djangocms_versioning').cms_extension
+    if versioning_extension.is_grouper_model_versioned(model):
+        return versioning_extension.versionables_by_grouper[model]
+
+
+@lru_cache()
```
czpython: woa. toolbar is a per request instance, you don't want this cached at the process level

djangocms_versioning/plugin_rendering.py
```diff
@@ -0,0 +1,54 @@
...
+            f for f in instance._meta.get_fields()
+            if f.is_relation and not f.auto_created
+        ]
+        for field in candidate_fields:
+            versionable = get_versionable_for_grouper(field.rel.model)
+            if not versionable:
+                continue
+            if self.toolbar.edit_mode_active or preview_mode_active(self.toolbar):
+                qs = versionable.content_model._base_manager.filter(
+                    versions__state__in=(DRAFT, PUBLISHED),
```
czpython: wouldn't this mean that for a single grouper we can get a draft version and a published version?

jonathan-s:
@czpython It would be great to know the reasoning behind why to use a published-only manager. I've had this break in unexpected ways because this is the default behaviour, so it would be nice to hear what this default behaviour solves.

PR #57, opened 2018-08-31
Implement correct edit link for objects with placeholders

Aiky30:
The edit redirect endpoint currently always redirects the user to the django admin change view. This is fine for objects which do not have placeholders (are not frontend-editable), but for objects with placeholders (such as page content objects) we should redirect to the frontend edit (the one with the cms toolbar).

Acceptance criteria:

The edit_redirect_view should redirect to the edit view inside the django admin for objects with no placeholders (same behaviour as now)

The edit_redirect_view should redirect to the edit view on the frontend (with the cms toolbar) for objects with placeholders


djangocms_versioning/admin.py
```diff
@@ -341,10 +344,15 @@ def edit_redirect_view(self, request, object_id):
         elif version.state != DRAFT:
             raise Http404
         # Redirect
-        url = reverse('admin:{app}_{model}_change'.format(
-            app=version.content._meta.app_label,
-            model=version.content._meta.model_name,
-        ), args=(version.content.pk,))
+        # If the object is editable the cms editable view should be used, with the toolbar.
+        if is_editable_model(version.content):
```
chaosk: `is_editable_model` takes a class, and although it seems to work for model instances when checking if model has a relation to Placeholder, it doesn't work with checking its admin for FrontendEditableAdminMixin inheritance.

PR #59, opened 2018-09-04
UI entry point into comparing versions (admin action)

Aiky30:
Create an admin action for comparing two versions (should redirect to the compare url). Has validation so that more than two versions cannot be selected.


djangocms_versioning/admin.py
```diff
@@ -198,6 +206,27 @@ def state_actions(self, obj):
...
     state_actions.short_description = 'actions'
 
+    def compare_versions_action(self, request, queryset):
+        """
+        Redirects to a compare versions view based on a users choice
+        """
+        # Validate that only two versions are selected
+        if queryset.count() is not 2:
+            # FIXME: Needs to be multilingual
+            messages.error(request, _("Two versions have to be selected."))
```
chaosk: Use `self.message_user`

djangocms_versioning/admin.py
```diff
@@ -198,6 +206,27 @@ def state_actions(self, obj):
...
+        # Build the link for the version comparison of the two selected versions
+        url = reverse('admin:{app}_{model}_compare'.format(
+            app=self.model._meta.app_label,
+            model=self.model._meta.model_name,
+        ), args=(queryset[0].pk,))
+        url += '?compare_to=%d' % queryset[1].pk
+
+        return redirect(url)
+
+    compare_versions_action.short_description = "Compare versions"
```
chaosk: Make it translatable

djangocms_versioning/admin.py
chaosk: Rephrase to say "exactly two versions"

PR #67, opened 2018-09-10
Remove explicit versionable parameter from filter_by_grouper

chaosk:


chaosk:
It's being removed because it's now possible to figure out the versionable based on grouper object itself. When filter_by_grouper was implemented there was no API for that, now there's no need to force users to provide it for us.

chaosk:
And yeah, PRs coming from outside of this repo aren't being built by CircleCI. Seems like a security measure (otherwise you could have anyone run code in build context of this project by submitting a PR)

PR #68, opened 2018-09-11
Use create_user instead of create in user factory to store hashed password

chaosk:


djangocms_versioning/test_utils/factories.py
```diff
@@ -25,6 +25,13 @@ class UserFactory(factory.django.DjangoModelFactory):
     class Meta:
         model = User
 
+    @classmethod
+    def _create(cls, model_class, *args, **kwargs):
```
Aiky30: I don't like overriding private methods but if this is the only way for overriding in the factory then it's ok.

PR #74, opened 2018-09-13
Version aware menu implementation

malindap92:
This PR consist of the implementation of the new menu implementation which is versioned aware. The implementation includes showing/hiding menu nodes on public mode, edit mode and preview mode.


djangocms_versioning/cms_config.py
```diff
@@ -157,6 +157,16 @@ def on_page_content_unpublish(version):
     page.clear_cache(menu=True)
 
 
+def on_page_content_draft_create(version):
+    page = version.content.page
+    page.clear_cache(menu=True)
+
+
+def on_page_content_archive(version):
```
dinolivanos: The body of this function is exactly the same as the one above. Is this right?

malindap92: yes it is but I kept it like that if in case we need to add extra kwargs. maybe can have _clear_menu_cache function and call it.

PR #85, opened 2018-09-19
Redirect to draft edit when attempting to edit published version when…

chaosk:
… there's already a draft version in place

### DIFF
--- a/djangocms_versioning/admin.py
+++ b/djangocms_versioning/admin.py
@@ -363,18 +363,10 @@ class VersionAdmin(admin.ModelAdmin):
         )) + '?grouper=' + str(version.grouper.pk)
         return redirect(url)
 
-    def edit_redirect_view(self, request, object_id):
-        """Redirects to the admin change view and creates a draft version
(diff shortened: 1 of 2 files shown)


djangocms_versioning/admin.py
```diff
@@ -363,18 +363,10 @@ def unpublish_view(self, request, object_id):
...
-        # This view always changes data so only POST requests should work
-        if request.method != 'POST':
-            return HttpResponseNotAllowed(['POST'], _('This view only supports POST method.'))
-
+    def _get_edit_redirect_version(self, request, object_id):
         version = self.get_object(request, unquote(object_id))
         if version is None:
-            return self._get_obj_does_not_exist_redirect(
-                request, self.model._meta, object_id)
+            return
```
Aiky30: This has lost the "_get_obj_does_not_exist_redirect" message. Obviously now this is a helper method for the view it can't just return the redirect here. Is it ok that it will always return a 404 rather than an admin error and message in this instance?

PR #88, opened 2018-09-20
Disable toolbar in compare view requests

chaosk:


tests/test_admin.py
```diff
@@ -1256,7 +1256,10 @@ def test_compare_view_has_version_data_in_context_when_no_get_param(self):
         self.assertIn('v1', context)
         self.assertEqual(context['v1'], versions[0])
         self.assertIn('v1_preview_url', context)
-        params = '?' + get_cms_setting('CMS_TOOLBAR_URL__PERSIST') + '=0'
+        params = (
```
Aiky30: I would put the param def as a property for the TestCase.
Example:
self.disable_toolbar_params = (
            get_cms_setting('CMS_TOOLBAR_URL__DISABLE') + '=1&' +
            get_cms_setting('CMS_TOOLBAR_URL__PERSIST') + '=0'
        )

then use it like so: 
my_url + '?' + self.disable_toolbar_params

I left the ? as a url may already define other params.

PR #92, opened 2018-09-21
emit change_page signal from cms_config

damilare:


djangocms_versioning/cms_config.py
```diff
@@ -141,7 +143,19 @@ def label_from_instance(obj, language):
     return "{title} ({path})".format(title=obj.get_title(language), path=obj.get_path(language))
 
 
+def emit_page_change(version):
+    # Trigger a post object operation
+    post_obj_operation.send(
+        sender=version.__class__,
+        operation=CHANGE_PAGE,
```
chaosk: Shouldn't this be CHANGE_PAGE_TRANSLATION?

damilare: Right, gotcha.

djangocms_versioning/cms_config.py
```diff
@@ -143,19 +143,25 @@ def label_from_instance(obj, language):
...
+    Sends a page content change signal for djangocms-internalsearch
+    if installed. It is used for re-indexing version state info
+    """
+    try:
+        from djangocms_internalsearch.signals import page_content_change_signal
+    except ImportError:
+        page_content_change_signal = None
+
+    if page_content_change_signal:
+        page_content_change_signal.send(
```
chaosk: Shouldn't it be a signal that handles all content objects?

djangocms_versioning/cms_config.py
```diff
@@ -143,19 +143,25 @@ def label_from_instance(obj, language):
...
-    )
+def emit_page_content_change(version):
+    """
+    Sends a page content change signal for djangocms-internalsearch
+    if installed. It is used for re-indexing version state info
+    """
+    try:
+        from djangocms_internalsearch.signals import page_content_change_signal
+    except ImportError:
+        page_content_change_signal = None
```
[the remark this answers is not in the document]
damilare: ok

djangocms_versioning/cms_config.py
damilare: True

djangocms_versioning/cms_config.py
```diff
@@ -147,11 +147,11 @@ def emit_page_content_change(version):
     if installed. It is used for re-indexing version state info
     """
     try:
```
chaosk: Perhaps do
```python
try:
    from djangocms_internalsearch.signals import content_object_state_change
except ImportError:
    def emit_page_content_change(version):
        pass
else:
    def emit_page_content_change(version):
        # implementation
        ...
```

damilare: one million ways to skin a cat, isn't your previous suggestion just fine?

damilare: I honestly find that easier to ready. Does this have a performance advantage?

chaosk: Doesn't it attempt to import every time the function is called?

PR #94, opened 2018-09-25
Grouper form choices fallback

chaosk:
Grouper form choices will now fallback to less important content objects using the following state order: published -> draft -> archived/unpublished

### DIFF
--- a/djangocms_versioning/datastructures.py
+++ b/djangocms_versioning/datastructures.py
@@ -1,6 +1,9 @@
 from django.core.exceptions import ImproperlyConfigured
-from django.db.models import Max
 from django.utils.functional import cached_property
+from django.db.models import Value as V, OuterRef, Prefetch, Subquery, Max
+
+from .constants import DRAFT, PUBLISHED
+from .compat import StrIndex
 
 from .models import Version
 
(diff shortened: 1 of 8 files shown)


djangocms_versioning/monkeypatch/page.py
```diff
@@ -18,6 +18,18 @@
 cms_extension = apps.get_app_config('cms').cms_extension
 
 
+def _get_title_cache(func):
```
Aiky30: I think we should also unit test this monkey patch.

djangocms_versioning/datastructures.py
```diff
@@ -73,6 +75,25 @@ def for_grouper(self, grouper):
...
+            order=StrIndex(
+                V(' '.join((DRAFT, PUBLISHED))),
+                'versions__state',
+            ),
+        ).filter(**{
+            self.grouper_field_name: OuterRef('pk'),
+        }).order_by('-order')
+        content_objects = self.content_model._base_manager.filter(
+            pk__in=self.grouper_model.objects.annotate(
+                content=Subquery(inner.values_list('pk')),
```
Aiky30: This will need to be fixed as per your findings to the error that I found.

PR #97, opened 2018-09-27
Emit content object change signal after every state change

chaosk:


djangocms_versioning/helpers.py
```diff
@@ -174,3 +174,27 @@ def version_list_url_for_grouper(grouper):
...
+    """
+    Sends a content change signal for djangocms-internalsearch
+    if installed. It is used for re-indexing version state info
+    """
+    try:
+        from djangocms_internalsearch.signals import content_object_state_change
+    except ImportError:
+        return
+
+    from djangocms_internalsearch.helpers import get_internalsearch_model_config
```
damilare: why not put this line in the try/except block

PR #98, opened 2018-09-27
Don't depend on get_internalsearch_model_config to check if IS is ena…

chaosk:
…bled and has content model registered with it


Aiky30:
The helpers no longer creates a breaking import, in fact the entire application is running and has been running fine in environments without internal search.

PR #100, opened 2018-09-27
Add support for extra grouping fields

chaosk:
Add filters for extra fields to version list
Add grouper to version list breadcrumbs


djangocms_versioning/admin.py
```diff
@@ -74,32 +73,61 @@ def change_view(self, request, object_id, form_url='', extra_context=None):
 class VersionChangeList(ChangeList):
 
     def get_filters_params(self, params=None):
+        content_model = self.model_admin.model._source_model
+        versionable = versionables.for_content(content_model)
         lookup_params = super().get_filters_params(params)
-        lookup_params.pop(GROUPER_PARAM, None)
+        lookup_params.pop(versionable.grouper_field_name, None)
```
mmoravcik: why not call this `filter_params`?

djangocms_versioning/admin.py
```diff
@@ -123,6 +151,13 @@ def get_queryset(self, request):
     def get_changelist(self, request, **kwargs):
         return VersionChangeList
 
+    def get_list_filter(self, request):
+        versionable = versionables.for_content(self.model._source_model)
+        return (
+            fake_filter_factory(versionable, field)
```
mmoravcik: Why `fake` in the name?
Also, if I wanted to filter by `created_by` too, how would I go about it?

djangocms_versioning/admin.py
```diff
@@ -467,13 +502,26 @@ def changelist_view(self, request, extra_context=None):
                 opts.app_label,
                 opts.model_name,
             )))
+        extra_context = extra_context or {}
+        versionable = versionables.for_content(self.model._source_model)
+        try:
+            grouper = versionable.get_grouper_with_fallbacks(
+                int(request.GET.get(versionable.grouper_field_name)),
+            )
+        except (ObjectDoesNotExist, TypeError, ValueError):
```
mmoravcik: ObjectDoesNotExist can't happen?

djangocms_versioning/admin.py
```diff
@@ -467,13 +502,26 @@ def changelist_view(self, request, extra_context=None):
...
+            grouper = None
+        extra_context.update(
```
mmoravcik: should this be in else?
try
catch
else

djangocms_versioning/admin.py
chaosk: Because it's not really filtering, docstring of `fake_filter_factory` says so.
I guess just add `('created_by', RelatedOnlyFieldListFilter)` there.

djangocms_versioning/admin.py
chaosk: You're correct

djangocms_versioning/admin.py
chaosk: `get_grouper_with_fallbacks` can return None anyway, so no point in that. I could add a `if grouper:`.

djangocms_versioning/admin.py
mmoravcik: Is there any documentation or explanation what `extra_grouping_fields` setting means? Or even better, or VersionableItem properties?

djangocms_versioning/datastructures.py
```diff
@@ -73,13 +77,34 @@ def distinct_groupers(self):
 
     def for_grouper(self, grouper):
         """Returns all `Content` objects for specified grouper object."""
-        return self.content_model._base_manager.filter(**{self.grouper_field.name: grouper})
+        return self.for_grouping_values(**{self.grouper_field.name: grouper})
+
+    def for_content_grouping_values(self, content):
+        return self.for_grouping_values(**self.grouping_values(content))
```
mmoravcik: I'd add docstrings to these methods?

djangocms_versioning/plugin_rendering.py
```diff
@@ -26,20 +26,21 @@ def render_plugin(self, instance, context, placeholder=None, editable=False):
...
                 qs = versionable.content_model.objects.all()
             related_field = getattr(instance, field.name)
             if related_field:
-                qs = qs.filter(**{
-                    versionable.grouper_field_name: related_field
-                })
+                filters = {
+                    versionable.grouper_field_name: related_field,
+                }
+                # TODO Figure out grouping values-awareness
```
mmoravcik: is there a story for this bit?

chaosk: There's no

mmoravcik: So the config variable is only valid for `language`? Should there be an exception if you try to pass anything else, or it is not necessary as it will be ignored in these filters?

PR #104, opened 2018-10-01
Added Revert action for Unpublished and Archived state

webbyfox:


djangocms_versioning/admin.py
```diff
@@ -249,12 +249,44 @@ def _get_edit_link(self, obj, request, disabled=False):
...
+        """Helper function to get the html link to the revert action
+        """
+        if obj.state in (UNPUBLISHED, ARCHIVED):
+            pks_for_grouper = obj.versionable.for_grouper(
+                obj.grouper).values_list('pk', flat=True)
+            drafts = Version.objects.filter(
+                object_id__in=pks_for_grouper, content_type=obj.content_type,
+                state=DRAFT)
+        else:
+            # Don't display  =the link if it's a draft or published
```
chaosk: Remove ` =`

djangocms_versioning/admin.py
```diff
@@ -249,12 +249,44 @@ def _get_edit_link(self, obj, request, disabled=False):
             }
         )
 
+    def _get_revert_link(self, obj, request, disabled=False):
+        """Helper function to get the html link to the revert action
+        """
+        if obj.state in (UNPUBLISHED, ARCHIVED):
+            pks_for_grouper = obj.versionable.for_grouper(
```
chaosk: Think of moving this out of per-row calculation. Currently this will run for every unpublished/archived version on the list. While the exact same code happens for published's the edit button, that doesn't happen as often (But could also be moved out of there).

djangocms_versioning/admin.py
```diff
@@ -249,12 +249,44 @@ def _get_edit_link(self, obj, request, disabled=False):
...
+            disable_revert_link = False
+            revert_url = reverse('admin:{app}_{model}_revert_redirect'.format(
+                app=obj._meta.app_label, model=self.model._meta.model_name,
+            ), args=(obj.pk,))
+
+        return render_to_string(
+            'djangocms_versioning/admin/revert_icon.html',
+            {
+                'revert_url': revert_url,
+                'disabled': disabled,
```
chaosk: What's the difference between disabled and disable_revert_link? Seems like it could be the same variable

djangocms_versioning/admin.py
```diff
@@ -412,6 +444,29 @@ def edit_redirect_view(self, request, object_id):
             ), args=(version.content.pk,))
         return redirect(url)
 
+    def revert_redirect_view(self, request, object_id):
+        """Redirects to the admin change view and creates a draft version
+        if no draft exists yet.
+        """
+        # This view always changes data so only POST requests should work
+        version = self.get_object(request, unquote(object_id))
+        version = version.copy(request.user)
```
chaosk: This happens too fast (before `version is None` check). Also you need to check if you're reverting from an archived/unpublished version, and that there's no existing drafts

djangocms_versioning/admin.py
```diff
@@ -412,6 +444,29 @@ def edit_redirect_view(self, request, object_id):
...
+
+        if version is None:
+            raise Http404
+
+        # Redirect
```
chaosk: Try to move this bit to a separate function

chaosk: Besides, didn't we think reverting should keep you on the version list?

djangocms_versioning/admin.py
```diff
@@ -412,6 +444,29 @@ def edit_redirect_view(self, request, object_id):
             ), args=(version.content.pk,))
         return redirect(url)
 
+    def revert_redirect_view(self, request, object_id):
```
chaosk: Rename to `revert_view` (same in url patterns)

djangocms_versioning/templates/djangocms_versioning/admin/revert_icon.html
```diff
@@ -0,0 +1 @@
+{% load static i18n %}<a {% if disable_revert_link %}class="btn cms-versioning-action-btn inactive" title="{% trans 'Draft already exists, revert action not possible' %}"{% else %}class="btn cms-versioning-action-btn js-versioning-action" href="{{ revert_url }}" title="{% trans 'Revert' %}"{% endif %}> <img src="{% static 'djangocms_versioning/svg/undo.svg' %}"></a>
```
chaosk: You're passing that `disabled` variable to the context, but it's not being used

chaosk: Also, you can split the tag into multiple lines so it's more readable (and then use `{% spaceless %}` around it, if you want to keep the output in one line)

djangocms_versioning/admin.py
```diff
@@ -259,23 +259,22 @@ def _get_revert_link(self, obj, request, disabled=False):
...
+            revert_url = reverse('admin:{app}_{model}_revert'.format(
                 app=obj._meta.app_label, model=self.model._meta.model_name,
             ), args=(obj.pk,))
 
         return render_to_string(
             'djangocms_versioning/admin/revert_icon.html',
             {
                 'revert_url': revert_url,
-                'disabled': disabled,
                 'disable_revert_link': disable_revert_link
```
chaosk: I'd rename it to `disabled`

djangocms_versioning/admin.py
```diff
@@ -433,38 +432,29 @@ def edit_redirect_view(self, request, object_id):
...
-        if is_editable_model(version.content.__class__):
-            url = get_object_edit_url(version.content)
-        # Or else, the standard edit view should be used
-        else:
-            url = reverse('admin:{app}_{model}_change'.format(
-                app=version.content._meta.app_label,
-                model=version.content._meta.model_name,
-            ), args=(version.content.pk,))
-        return redirect(url)
+        return redirect(get_editable_url(version))
```
chaosk: Perhaps change it so `get_editable_url` expects a content object? You're not using any other version property than `version.content` inside

djangocms_versioning/admin.py
```diff
@@ -433,38 +432,29 @@ def edit_redirect_view(self, request, object_id):
...
             raise Http404
 
+        pks_for_grouper = version.versionable.for_grouper(
+            version.grouper).values_list('pk', flat=True)
+
+        drafts = Version.objects.filter(
+            object_id__in=pks_for_grouper, content_type=version.content_type,
+            state=DRAFT)
+
+        if not drafts.exists() and version.state in (UNPUBLISHED, ARCHIVED):
```
chaosk: Add more error handling, currently you'll redirect to old version if there's an existing draft.

djangocms_versioning/static/djangocms_versioning/js/actions.js
```diff
@@ -19,12 +19,21 @@
                         '">' +
                     '</form>'
                 );
-
+                keepSideFrame = action.attr('class').indexOf('js-versioning-keep-sideframe') === -1
```
chaosk: Add `var`

chaosk: Also, the name suggests value is true when you want to keep the sideframe, but the value is inverted.

chaosk: It should be `!== 1`

djangocms_versioning/static/djangocms_versioning/js/actions.js
```diff
@@ -19,12 +19,21 @@
...
                 // always break out of the sideframe, cause it was never meant to open cms views inside it
                 try {
-                    window.top.CMS.API.Sideframe.close();
+                    if (keepSideFrame)
```
chaosk: `!keepSideFrame`

djangocms_versioning/templates/djangocms_versioning/admin/revert_icon.html
```diff
@@ -0,0 +1,9 @@
+{% load static i18n %}
+{% spaceless %}
+<a
+{% if disable %} class="btn cms-versioning-action-btn js-versioning-keep-sideframe inactive" title="{% trans 'Draft already exists, revert action not possible' %}"
+{% else %} class="btn cms-versioning-action-btn js-versioning-action js-versioning-keep-sideframe" href="{{ revert_url }}" title="{% trans 'Revert' %}"{% endif %}>
+ <img src="{% static 'djangocms_versioning/svg/undo.svg' %}">
+</a>
+{% endspaceless %}
+
```
chaosk: Remove that last newline

PR #106, opened 2018-10-03
CMS default edit button override

Aiky30:


djangocms_versioning/cms_toolbars.py
```diff
@@ -86,7 +86,21 @@ def _add_edit_button(self, disabled=False):
...
+
+            # Try and locate an edit button
+            found = [button for button in button_list.buttons if button.name == _('Edit')]
+            if found:
+                button_list.buttons.remove(found[0])
+                return
+
     def post_template_populate(self):
         super(VersioningToolbar, self).post_template_populate()
+        self._remove_cms_edit_button()
```
malindap92: I don't think you will need to iterate button items and remove the Edit button. With the new implementation in CMS 4.0.x, you can extend VersioningToolbar with PlaceholderToolbar and override `add_edit_button`.

Aiky30: @malindap92 I couldn't remember the plan for this, extending the Placeholder forces a page object which felt wrong, although I agree that targeting by the label "edit" is poor. Can you remember if the plan was to extend the Placeholder toolbar?

malindap92: Earlier the buttons were rendered by PageToolbar. This has been moved to PlaceholderToolbar and its implemented in a way where you can extend and override `add_edit_button` and `add_preview_button`. Please do clarify with @chaosk as well.

chaosk: @malindap92 Soo, we need to inherit from PlaceholderToolbar, unregister it and register our toolbar, right? I think @Aiky30 raised a point that some other apps could be trying to do the same, and once PlaceholderToolbar is unregistered they wouldn't be able to find ours.

malindap92: @chaosk Yes that was my assumption that VersioningToolbar will override PlaceholderToolbar. In ModerationToolbar, it inherits VersioningToolbar and unregister it. Version Locking monkey patches the VersioningToolbar. Please do point me if I'm missing something here.

Aiky30: If the buttons where removed from the PageToolbar to facilitate overriding the PlaceholderToolbar we should go with that rather than monkey patching or as I've done look by the button name.

Aiky30:
@chaosk @malindap92 This is now passing and ready for what is hopefully a final review.

djangocms_versioning/cms_toolbars.py
```diff
@@ -67,29 +66,32 @@ def _add_publish_button(self):
...
             self.toolbar.add_item(item)
 
+    def add_edit_button(self):
+        """
+        Only override the CMS versioning button when the object is versionable
+        """
+        if not self._is_versioned():
+            # Show the standard cms edit button for non versionable objects
+            return super(VersioningToolbar, self).add_edit_button()
+        return self._add_edit_button()
```
malindap92: no need of return here

tests/test_toolbars.py
```diff
@@ -19,11 +25,16 @@ def _get_toolbar(self, content_obj, **kwargs):
...
         request.session = {}
-        cms_toolbar = CMSToolbar(request)
+        request.current_page = None if not hasattr(content_obj, 'page') else content_obj.page
+        request.toolbar = CMSToolbar(request)
         toolbar = VersioningToolbar(
-            request, toolbar=cms_toolbar, is_current_app=True, app_path='/')
+            request,
+            toolbar=request.toolbar,
+            is_current_app=True,
+            app_path='/'
```
malindap92: trailing comma

djangocms_versioning/cms_toolbars.py
```diff
@@ -67,29 +66,32 @@ def _add_publish_button(self):
             )
             self.toolbar.add_item(item)
 
+    def add_edit_button(self):
+        """
+        Only override the CMS versioning button when the object is versionable
+        """
+        if not self._is_versioned():
+            # Show the standard cms edit button for non versionable objects
+            return super(VersioningToolbar, self).add_edit_button()
```
chaosk: Change it to `super()`

tests/test_toolbars.py
```diff
@@ -19,11 +25,21 @@ def _get_toolbar(self, content_obj, **kwargs):
         """Helper method to set up the toolbar
         """
         request = RequestFactory().get('/')
-        request.user = UserFactory()
+        request.user = self.get_superuser()
         request.session = {}
-        cms_toolbar = CMSToolbar(request)
-        toolbar = VersioningToolbar(
-            request, toolbar=cms_toolbar, is_current_app=True, app_path='/')
+        request.current_page = None if not hasattr(content_obj, 'page') else content_obj.page
```
chaosk: `getattr(content_obj, 'page', None)`?

PR #136, opened 2018-10-24
Move can_* from admin class to Version model

chaosk:


djangocms_versioning/admin.py
```diff
@@ -576,9 +568,8 @@ def revert_view(self, request, object_id):
         if version is None:
             raise Http404
 
-        if version.state not in (UNPUBLISHED, ARCHIVED):
-            # if version state not unpublished or archived then raise 404
-            raise Http404
+        if not version.can_revert(request.user):
```
webbyfox: `or not version.can_revert(request.user)` ?

PR #146, opened 2018-10-30
Fix PageContent admin change view error when request.user cannot modify this version

toffi9:


djangocms_versioning/cms_config.py
```diff
@@ -9,6 +10,9 @@
 from cms.models import PageContent, Placeholder
 from cms.utils.i18n import get_language_tuple
 
+from djangocms_versioning.admin import VersioningAdminMixin
+from djangocms_versioning.models import Version
+
```
chaosk: Use relative imports

PR #153, opened 2018-11-26
Added version number and state to the menu label

Aiky30:


djangocms_versioning/cms_toolbars.py
```diff
@@ -106,8 +106,12 @@ def _add_versioning_menu(self):
         if version is None:
             return
 
+        version_menu_label = _('Version {number} {state}').format(
```
chaosk:
```suggestion
        version_menu_label = _('Version #{number} ({state})').format(
```
What do you think?

PR #167, opened 2019-02-05
Updates modified date when version state changes

webbyfox:


tests/test_admin.py
```diff
@@ -847,6 +847,27 @@ class ArchiveViewTestCase(BaseStateTestCase):
...
+        url = self.get_admin_url(
+            self.versionable.version_model_proxy, 'archive', poll_version.pk)
+
+        user = self.get_staff_user_with_no_permissions()
+
+        with self.login_user_context(user):
+            self.client.post(url)
+
+        # get object again after state change
+        poll_version_ = Version.objects.get(pk=poll_version.pk)
```
monikasulik: Why not just `poll_version.refresh_from_db()`?

tests/test_admin.py
```diff
@@ -847,6 +847,27 @@ class ArchiveViewTestCase(BaseStateTestCase):
...
+        # check time is not same as before
+        self.assertNotEqual(poll_version_.modified, poll_version.modified)
```
monikasulik: What are you achieving in this assert that you're not achieving with the assert below? To me it looks suspiciously like you're testing if you're mocking worked rather than whether the field is being updated correctly. If you do the mocking of this with freezegun you can be confident it's being done ok :)

tests/test_admin.py
```diff
@@ -1184,6 +1223,22 @@ def test_unpublish_view_sets_state_and_redirects(self, mocked_messages):
         # Redirect happened
         self.assertRedirectsToVersionList(response, poll_version)
 
+    @freeze_time("2099-01-11")
```
monikasulik: and ditto

tests/test_admin.py
```diff
@@ -1282,6 +1337,30 @@ def test_unpublish_view_can_be_accessed_by_get_request(self):
         self.assertEqual(StateTracking.objects.all().count(), 0)
 
 
+class RevertViewTestCase(BaseStateTestCase):
+
+    def setUp(self):
+        self.versionable = PollsCMSConfig.versioning[0]
+
+    @freeze_time("2099-01-11")
```
monikasulik: and ditto

PR #168, opened 2019-02-07
Add config option to add to unpublish confirmation context

monikasulik:


chaosk:
How's this going to be used by the unpublish view?

docs/versioning_integration.md
```diff
@@ -73,6 +77,9 @@ class BlogCMSConfig(CMSAppConfig):
             preview_url=get_preview_url
         ),
     ]
+    versioning_add_to_confirmation_context = {
+        'unpublish': [stories_about_intelligent_cats],
```
monikasulik: @chaosk How would you feel about changing this to `'unpublish': {'cats': stories_about_intelligent_cats, 'dogs': dog_stories}`?
This would make the context dict something like `{..., 'extra_context': {'cats': ..., 'dogs': ...}}` and the template:
```
{% for thing in extra_context.values %}
    {{thing}}
{% endfor %}
```
But would mean that a project could override that template and reference vars like `{{extra_context.cats}}` and `{{extra_context.dogs}}`

Or we can leave this as a list as it is now.

chaosk: Am still thinking if this is the correct approach, or should that function return a rendered template?

But yes, changing the structure to a dict makes sense

tests/test_cms_config.py
```diff
@@ -123,10 +162,66 @@ def test_versionables_list_created(self):
...
             extension.versionables, [poll_versionable, blog_versionable])
 
+    def test_context_dict_created(self):
+        """Test a dict is populated from the versioning_add_to_confirmation_context
+        config
+        """
+        extension = VersioningCMSExtension()
+
+        def unpublish_context1(request, version, *args, **kwargs):
+            return "Every time you unpublish something you should kiss a cat"
```
webbyfox: Do we need to return if you are not testing it? `pass` should do the job. Although the message does spice up reading the test.

tests/test_cms_config.py
```diff
@@ -123,10 +162,66 @@ def test_versionables_list_created(self):
...
+
+        def unpublish_context2(request, version, *args, **kwargs):
+            return "Good luck with your unpublishing"
```
webbyfox: same as above

tests/test_cms_config.py
```diff
@@ -123,10 +162,66 @@ def test_versionables_list_created(self):
...
+            extension.add_to_context, {'unpublish': [unpublish_context1, unpublish_context2]})
+
+    def test_context_dict_doesnt_get_overwritten(self):
+        """Test when multiple apps update the same key in the context dict,
+        values from both apps end up under that key
+        """
+        extension = VersioningCMSExtension()
+
+        def unpublish_context1(request, version, *args, **kwargs):
+            return "Cats don't like to be unpublished"
```
webbyfox: same as above

tests/test_cms_config.py
```diff
@@ -123,10 +162,66 @@ def test_versionables_list_created(self):
...
+
+        def unpublish_context2(request, version, *args, **kwargs):
+            return "Elephants don't mind being unpublished'"
```
webbyfox: same as above

tests/test_cms_config.py
monikasulik: These messages are asserted later in the test, so I do actually need them... :cat2: :mouse2:

monikasulik: Hmmm, although actually not in that test :D I guess I could put `pass`, but I think I prefer to return some sort of string as that's what a real function would do (but I can make the message less spicy if you think could be a problem ;)).

monikasulik: @chaosk Any opinions on this?

chaosk: I don't have a preference here, you can keep it as it is

djangocms_versioning/cms_config.py
```diff
@@ -73,6 +71,22 @@ def handle_versioning_setting(self, cms_config):
...
+        """
+        add_to_context = cms_config.versioning_add_to_confirmation_context
+        supported_keys = ['unpublish']
+        for key, value in add_to_context.items():
+            if key not in supported_keys:
+                raise ImproperlyConfigured(
+                    "{!r} is not a supported dict key in the versioning_add_to_confirmation_context setting".format(key)
+                )
+            if key not in self.add_to_context:
+                self.add_to_context[key] = {}
```
monikasulik: This needs to be an OrderedDict as well.

djangocms_versioning/cms_config.py
```diff
@@ -25,6 +25,7 @@ class VersioningCMSExtension(CMSAppExtension):
 
     def __init__(self):
         self.versionables = []
+        self.add_to_context = collections.OrderedDict()
```
monikasulik: On second thought, this probably doesn't need to be an OrderedDict as we only ever take one element of this dict per view.

docs/versioning_integration.md
```diff
@@ -89,6 +97,21 @@ class BlogCMSConfig(CMSAppConfig):
     - preview_url - This is optional attribute can be pass to override preview url for an object in version list
     table. If it is not passed then if model is a editable, it will render object preview url else
     changelist url.
+3. The `versioning_add_to_confirmation_context` is a OrderDict where the keys are
```
monikasulik: typo: an OrderedDict

tests/test_admin.py
```diff
@@ -1481,6 +1483,55 @@ def test_unpublish_view_can_be_accessed_by_get_request(self):
...
+
+        def unpublish_context2(request, version, *args, **kwargs):
+            return "Unpublish the mice instead."
+
+        def publish_context(request, version, *args, **kwargs):
+            return "Publish cat pictures only. People aren't interested in anything else."
+
+        versioning_ext = apps.get_app_config('djangocms_versioning').cms_extension
+        extra_context = OrderedDict({
+            'unpublish': {'cats': unpublish_context1, 'mice': unpublish_context2},
```
monikasulik: The other way round, it's the dicts inside which need to be OrderedDict objects.

docs/versioning_integration.md
```diff
@@ -73,6 +78,9 @@ class BlogCMSConfig(CMSAppConfig):
             preview_url=get_preview_url
         ),
     ]
+    versioning_add_to_confirmation_context = OrderDict({
+        'unpublish': {'cat_stories': stories_about_intelligent_cats},
```
monikasulik: This should be the other way round. It's the inner dict that needs to be ordered.

tests/test_cms_config.py
```diff
@@ -107,6 +109,46 @@ def test_raises_exception_if_content_class_already_registered_in_different_confi
...
+            copy_function=default_copy,
+        )
+        cms_config = Mock(
+            spec=[],
+            djangocms_versioning_enabled=True,
+            versioning=[poll_versionable],
+            # versioning doesn't know what red rabbits is
+            # so this should raise an exception
+            versioning_add_to_confirmation_context={
+                'red_rabbits': {'rabbit': lambda r, v: v.content}
```
monikasulik: should use OrderedDict

djangocms_versioning/cms_config.py
```diff
@@ -73,6 +71,22 @@ def handle_versioning_setting(self, cms_config):
...
+        and add it to the master dict if all is ok
+        """
+        add_to_context = cms_config.versioning_add_to_confirmation_context
+        supported_keys = ['unpublish']
+        for key, value in add_to_context.items():
+            if key not in supported_keys:
+                raise ImproperlyConfigured(
+                    "{!r} is not a supported dict key in the versioning_add_to_confirmation_context setting".format(key)
+                )
+            if key not in self.add_to_context:
```
chaosk: Can be removed if `self.add_to_context = {}` gets changed to `self.add_to_context = defaultdict(OrderedDict)``

PR #172, opened 2019-02-13
Versioning publishing signals

Aiky30:
Make versioning emit signals for publish, unpublish and archive.

### DIFF
--- a/djangocms_versioning/models.py
+++ b/djangocms_versioning/models.py
@@ -9,6 +9,7 @@ from django_fsm import FSMField, can_proceed, transition
 
 from . import constants, versionables
 from .conditions import Conditions, in_state
+from .operations import send_post_version_operation, send_pre_version_operation
 
 
 try:
@@ -102,13 +103,12 @@ class Version(models.Model):
         return "Version #{}".format(self.pk)
 
     def save(self, **kwargs):
-        from .operations import send_post_version_operation, send_pre_version_operation
-
         created = not self.pk
         # On version creation
         if created:
             # trigger pre operation signal
-            action_token = send_pre_version_operation(constants.OPERATION_DRAFT, obj=self)
+            action_token = send_pre_version_operation(sender=self.content_type,operation=constants.OPERATION_DRAFT,
+                                                      obj=self)
             # Set the version number
             self.number = self.make_version_number()
 
@@ -128,7 +128,8 @@ class Version(models.Model):
                 if on_draft_create:
                     on_draft_create(self)
                 # trigger post operation signal
-                send_post_version_operation(constants.OPERATION_DRAFT, action_token, obj=self)
+                send_post_version_operation(sender=self.content_type, operation=constants.OPERATION_DRAFT,
+                                            token=action_token, obj=self)
             if emit_content_change:
                 emit_content_change(self.content, created=created)
 
@@ -180,10 +181,9 @@ class Version(models.Model):
         return can_proceed(self._set_archive)
 
     def archive(self, user):
-        from .operations import send_post_version_operation, send_pre_version_operation
-
         # trigger pre operation signal
-        action_token = send_pre_version_operation(constants.OPERATION_ARCHIVE, obj=self)
+        action_token = send_pre_version_operation(sender=self.content_type, operation=constants.OPERATION_ARCHIVE,
+                                                  obj=self)
         """Change state to ARCHIVED"""
         self._set_archive(user)
         self.modified = timezone.now()
@@ -198,7 +198,8 @@ class Version(models.Model):
         if on_archive:
             on_archive(self)
         # trigger post operation signal
-        send_post_version_operation(constants.OPERATION_ARCHIVE, action_token, obj=self)
+        send_post_version_operation(sender=self.content_type, operation=constants.OPERATION_ARCHIVE,
+                                    token=action_token, obj=self)
         if emit_content_change:
             emit_content_change(self.content)
 
@@ -225,10 +226,9 @@ class Version(models.Model):
     def publish(self, user):
         """Change state to PUBLISHED and unpublish currently
         published versions"""
-        from .operations import send_post_version_operation, send_pre_version_operation
-
         # trigger pre operation signal
-        action_token = send_pre_version_operation(constants.OPERATION_PUBLISH, obj=self)
+        action_token = send_pre_version_operation(sender=self.content_type, operation=constants.OPERATION_PUBLISH,
+                                                  obj=self)
         self._set_publish(user)
         self.modified = timezone.now()
         self.save()
@@ -251,7 +251,8 @@ class Version(models.Model):
         if on_publish:
             on_publish(self)
         # trigger post operation signal
-        send_post_version_operation(constants.OPERATION_PUBLISH, action_token, obj=self)
+        send_post_version_operation(sender=self.content_type, operation=constants.OPERATION_PUBLISH,
+                                    token=action_token, obj=self)
         if emit_content_change:
             emit_content_change(self.content)
 
@@ -276,10 +277,9 @@ class Version(models.Model):
         return can_proceed(self._set_unpublish)
 
     def unpublish(self, user):
-        from .operations import send_post_version_operation, send_pre_version_operation
-
         # trigger pre operation signal
-        action_token = send_pre_version_operation(constants.OPERATION_UNPUBLISH, obj=self)
+        action_token = send_pre_version_operation(sender=self.content_type, operation=constants.OPERATION_UNPUBLISH,
+                                                  obj=self)
         """Change state to UNPUBLISHED"""
         self._set_unpublish(user)
         self.modified = timezone.now()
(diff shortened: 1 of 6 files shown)


djangocms_versioning/models.py
```diff
@@ -102,9 +102,13 @@ def __str__(self):
         return "Version #{}".format(self.pk)
 
     def save(self, **kwargs):
+        from .operations import send_post_version_operation, send_pre_version_operation
+
         created = not self.pk
         # On version creation
         if created:
+            # trigger pre operation signal
+            action_token = send_pre_version_operation(constants.DRAFT, obj=self)
```
chaosk: Think we've discussed that this signal (on version creation) is not needed, because pre/post_save signals are already there

djangocms_versioning/models.py
```diff
@@ -174,6 +180,10 @@ def can_be_archived(self):
         return can_proceed(self._set_archive)
 
     def archive(self, user):
+        from .operations import send_post_version_operation, send_pre_version_operation
+
+        # trigger pre operation signal
+        action_token = send_pre_version_operation(constants.ARCHIVED, obj=self)
```
chaosk: Please decouple operation name from version state constants (use another constant, for example `constants.OPERATION_ARCHIVE` instead of `constants.ARCHIVE`)

djangocms_versioning/models.py
Aiky30: I would argue that the exisiing standard signal should be disabled and this custom one is kept as it keeps them consistent (draft would be different) and makes them easier to reuse.

monikasulik: Disabling standard django pre_save and post_save signals is likely to not be straightforward. In terms of keeping consistency I think it's more important to keep things consistent with django and django-cms (which doesn't disable any standard django signals AFAIK) than with versioning which is a library most developers will know less about than django and django-cms.

Aiky30: @monikasulik I agree to not trying to remove them as I originally said. I don't agree with using the Django default signals just for the DRAFT creation as they will contain a slightly different structure and will be different compared to the Versioning supplied ones other than DRAFT i.e. ARCHIVE, PUBLISH, UNPUBLISH.

What are your thoughts?

tests/test_signals.py
```diff
@@ -0,0 +1,114 @@
...
+            self.assertEqual(post_call_kwargs['operation'], constants.OPERATION_ARCHIVE)
+            self.assertEqual(pre_call_kwargs['obj'], version)
+            # post call
+            self.assertTrue('token' in post_call_kwargs)
+            self.assertEqual(post_call_kwargs['operation'], constants.OPERATION_ARCHIVE)
+            self.assertEqual(post_call_kwargs['obj'], version)
+
+    def test_draft_signals_fired(self):
+        """
+        When a version is changed to draft the correct signals are fired!
```
monikasulik: The phrasing in this docstring is technically wrong as it's impossible to actually "change a version to draft". A version starts off its life as a draft, but at no point is it possible to change a version to be a draft. Which is one of the reasons I agree with Krzysztof on this one - I don't really see having a signal for draft as "adding consistency", on the contrary.

djangocms_versioning/models.py
monikasulik: My thoughts atm are to follow YAGNI :D https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it I do feel draft is a different case as we never actually "change" to it. But maybe if you explain how you intend to be using these signals in the migration project I will change my mind ;)

djangocms_versioning/operations.py
```diff
@@ -0,0 +1,25 @@
+# -*- coding: utf-8 -*-
+import uuid
+
+from .models import Version
+from .signals import post_version_operation, pre_version_operation
+
+
+def send_pre_version_operation(operation, sender=Version, **kwargs):
+    token = str(uuid.uuid4())
```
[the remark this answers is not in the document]
chaosk: @monikasulik Think this is similar to how CMS signals work, where you can tie pre/post signals using that token

djangocms_versioning/operations.py
```diff
@@ -0,0 +1,25 @@
...
+        operation=operation,
+        token=token,
+        **kwargs
+    )
+    return token
+
+
+def send_post_version_operation(operation, token, sender=Version, **kwargs):
+    post_version_operation.send(
+        sender=sender,
```
chaosk: `sender=obj.content_type.model_class()`

chaosk:
Still need the docs for the signals

Aiky30:
@chaosk Docs in the code, PR or in the docs Noel is managing?

tests/test_signals.py
Aiky30: Let's just say I have a feeling that firing a signal on Draft will be useful. Every time I have stood by something and the team push back it's later been proven that it would have been useful.

chaosk:
Add them to the docs folder in the repo, we'll move them to sphinx-based format later on

djangocms_versioning/models.py
Aiky30: That's a terrible principle. Don't implement performance increases until you need them, in which case it's much harder to refactor rather than bake it in.

monikasulik: Actually, the one time I was on a project that religiously followed YAGNI, it was the easiest project I've ever had to maintain. And that despite the fact that it had a very similar architecture to what we've been doing on gaps (i.e. a constantly growing number of repos that have to work together).
Besides, performance issues are much more likely to happen when something is over-engineered than under-engineered. And when a performance issue does happen in code that's under-engineered it tends to be much easier to correct (it'll be something like adding select_related somewhere vs. spending hours figuring out where exactly an issue is happening in lots of code that was made more abstract that it needed to be).
And also, if you know you'll need the performance increase (as we do when working on gaps) then it's not breaking YAGNI to have it in.

docs/signals.md
```diff
@@ -0,0 +1,32 @@
...
+    - When a version is created the operation sent is 'operation_draft'
+    - When a version is archived the operation sent is 'operation_archive'
+    - When a version is published the operation emitted is 'operation_publish'
+    - When a version is un-published the operation emitted is 'operation_unpublish'
+
+## How To
+
+### Using the version publish and un-publish signal for a CMS Page
+
+The CMS used to provide page publish and unpublish signals which have since been removed. To replicate the behaviour you can listen to changes on the cms model PageContent as shown in the example below.
```
monikasulik: I think it would be better to be more specific here - i.e. specify that the CMS provided these signals before version 4.0.

docs/signals.md
```diff
@@ -0,0 +1,32 @@
...
+
+
+@receiver(post_version_operation, sender=PageContent)
+def do_something_on_page_publish_unpublsh(*args, **kwargs):
+
+    if (kwargs['operation'] == constants.OPERATION_PUBLISH or
+       kwargs['operation'] == constants.OPERATION_UNPUBLISH):
+        # ... do something 
+```
+
```
monikasulik: These docs are missing information on what the tokens are and why one would use them.

tests/test_signals.py
```diff
@@ -120,3 +123,28 @@ def test_draft_signals_fired(self):
...
+
+        version_1 = factories.PageVersionFactory(state=constants.DRAFT, content__template="")
+        version_2 = factories.PageVersionFactory(state=constants.DRAFT, content__template="")
+        version_1.publish(self.superuser)
+        version_1.unpublish(self.superuser)
+        version_2.archive(self.superuser)
+
+        # Only the publish and unpublish signals should have had an affect
+        self.assertEqual(len(signal_hits), 2)
+        self.assertEqual(signal_hits[0].state, constants.UNPUBLISHED)
```
monikasulik: Shouldn't this test fail on this line? The first thing you do with the version is publish it, so why would the constant be UNPUBLISHED?

djangocms_versioning/models.py
```diff
@@ -174,6 +179,8 @@ def can_be_archived(self):
         return can_proceed(self._set_archive)
 
     def archive(self, user):
+        # trigger pre operation signal
+        action_token = send_pre_version_operation(constants.OPERATION_ARCHIVE, version=self)
```
monikasulik: Why are you putting this ahead of the docstring? The docstring should be the first thing in a function, surely?

djangocms_versioning/models.py
```diff
@@ -258,6 +271,8 @@ def can_be_unpublished(self):
         return can_proceed(self._set_unpublish)
 
     def unpublish(self, user):
+        # trigger pre operation signal
+        action_token = send_pre_version_operation(constants.OPERATION_UNPUBLISH, version=self)
```
monikasulik: Again you're putting this ahead of the docstring - any reason for this?

djangocms_versioning/operations.py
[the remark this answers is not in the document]
monikasulik: Well, I'm not familiar with how they used to work and since from 4.0 they won't be there, I still think this should be documented and explained :) Not even sure what "tying" a signal means.

tests/test_signals.py
```diff
@@ -146,5 +149,5 @@ def do_something_on_page_publish_unpublsh(*args, **kwargs):
 
         # Only the publish and unpublish signals should have had an affect
         self.assertEqual(len(signal_hits), 2)
-        self.assertEqual(signal_hits[0].state, constants.UNPUBLISHED)
-        self.assertEqual(signal_hits[1].state, constants.UNPUBLISHED)
+        self.assertEqual(signal_hits[0].get('state'), constants.PUBLISHED)
```
monikasulik: Why `.get('state')` and not just `['state']`? In a test it's usually better to be as specific as possible. If things break in a test, it's best if they break as loudly as possible.

Aiky30: .get('state') will return None which would fail the check.

PR #175, opened 2019-02-20
Make checks function work with unversioned content

chaosk:


djangocms_versioning/helpers.py
```diff
@@ -193,12 +193,18 @@ def version_list_url_for_grouper(grouper):
 
 
 def is_content_editable(placeholder, user):
-    """A helper method for monkey patch to check version is in edit state
+    """A helper method for monkey patch to check version is in edit state.
+    Returns True if placeholder is related to a source object
+    which not versioned.
```
monikasulik: Typo: which *is* not versioned

PR #177, opened 2019-02-21
Better version admin breadcrumbs

monikasulik:


djangocms_versioning/admin.py
```diff
@@ -753,6 +753,24 @@ def changelist_view(self, request, extra_context=None):
                 grouper=grouper,
                 title=_('Displaying versions of "{grouper}"').format(grouper=grouper),
             )
+            grouping_values = {
+                field: request.GET.get(field)
+                for field in versionable.extra_grouping_fields
+                if request.GET.get(field) is not None
+            }
+            grouping_values[versionable.grouper_field_name] = grouper
+            extra_context['latest_content'] = Version.objects.filter_by_grouping_values(
```
chaosk: Think this approach is fine, but this filtering is already done in ChangeList, is there no way to grab that?

djangocms_versioning/templates/admin/djangocms_versioning/change_list.html
```diff
@@ -2,12 +2,5 @@
 {% load i18n admin_urls %}
```
webbyfox: You can get away probably by not loading `i18n` here as not needed in this template.

djangocms_versioning/admin.py
monikasulik: I'll look into it.

PR #178, opened 2019-02-28
Added sphinx setup and reformatted existing docs to rst format

webbyfox:
To install sphinx:
Run:
`pip install -U Sphinx`
Then cd `docs/ ` and run 
`make html`

which should generate a `_build` folder with HTML of documentation which can be browsed.


docs/signals.rst
```diff
@@ -9,26 +13,26 @@ Signals are fired before and after the following events which can be found in th
 
 A token is emitted in the signals that will allow the pre and post signals to be tied together, this could be of use if multiple transactions occur at the same time, allowing a token to match the pre and post signals that belong together.
 
-## How To
-
-### Using the version publish and un-publish signal for a CMS Page
+How To Using the version publish and un-publish signal for a CMS Page
```
monikasulik: "How to Use" not "How To Using"

PR #181, opened 2019-03-24
Added new parameter for filtering in distinct_groupers func

webbyfox:


tests/test_datastructures.py
```diff
@@ -38,6 +39,28 @@ def test_distinct_groupers(self):
             ordered=False,
         )
 
+    def test_queryset_filter_for_distinct_groupers(self):
+        poll1_version = PollVersionFactory(
+            content__poll=self.initial_version.content.poll, state=PUBLISHED
+        )
+        poll2_version = PollVersionFactory()
+        PollVersionFactory(content__poll=poll2_version.content.poll)
+        PollVersionFactory(content__poll=poll2_version.content.poll)
```
monikasulik: I'm not sure this data is adequate to really test this properly tbh. I mean, for example if you had a bug where the `distinct_groupers` method didn't have distinct groupers when filtering by additional params, I don't think this test would catch this.

Also, I'm wondering what happens in a situation like so...
Lets say you had these 4 objects in the db:
1 = poll1_grouper, archived
2 = poll1_grouper, published
3 = poll2_grouper, archived
4 = poll2_grouper, archived
And then you did `versionable.distinct_groupers(versions__state__in=ARCHIVED)`
What would happen? Obviously 2 would not be in the queryset, but what about 3 and 4? I guess only one of those would be in the queryset, but which one?

webbyfox: Tbh, When there is a bug in `distinct_groupers` for distinct groupers that should be cover in outside this test. I added the filtering parameter which I tried to cover here.

Regarding the situation, you have mentioned. there should be two latest ARCHIVED objects in queryset.

djangocms_versioning/datastructures.py
```diff
@@ -78,17 +78,16 @@ def version_model_proxy(self):
     def grouper_model(self):
         return self.grouper_field.remote_field.model
 
-    def distinct_groupers(self):
+    def distinct_groupers(self, **kwargs):
         """Returns a queryset of `self.content` objects with unique
         grouper objects.
 
         Useful for listing, e.g. all Polls.
```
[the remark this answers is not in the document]
monikasulik: We're using the sphinx formatting for params in various places AFAIK: https://pythonhosted.org/an_example_pypi_project/sphinx.html#function-definitions So would be better to stick to that.

djangocms_versioning/datastructures.py
```diff
@@ -78,17 +78,20 @@ def version_model_proxy(self):
...
+
+        Args:
+            kwargs: Optionally it takes dict object (**kwargs)
```
chaosk: Please improve docstring here. Something like "optional filtering parameters for inner queryset". Also the format is wrong, we aren't using google-style param formatting

PR #183, opened 2019-03-25
Add permission handling to version list, only show toolbar entry if user has permission to view it

chaosk:


djangocms_versioning/models.py
```diff
@@ -151,6 +153,14 @@ def versionable(self):
         """
         return versionables.for_content(self.content)
 
+    def convert_to_proxy(self):
```
monikasulik: I think this should have a test.

djangocms_versioning/admin.py
```diff
@@ -826,7 +826,9 @@ def has_add_permission(self, request):
     def has_change_permission(self, request, obj=None):
         """Return True for changelist and False for change view.
         """
-        return obj is None
+        if obj is not None:
```
monikasulik: Don't see a test for this change and yet, assuming the above comment is True, it did change the permission logic.

monikasulik:
Also, conflicts and flake8 seems to be failing.

tests/test_admin.py
```diff
@@ -2119,6 +2119,35 @@ def test_changelist_view_redirects_on_url_params_that_dont_exist(self):
         self.assertEqual(response.status_code, 302)
         self.assertEqual(response.url, expected_redirect)
 
+    def test_changelist_view_requires_change_permission(self):
```
monikasulik: Is there a test for when the permission has been added? I think all the other changelist views work on superuser, so there should probably be a test that checks this for a regular staff user with the change permission.

PR #190, opened 2019-09-12
Allow draft to be accessible from monkeypatched queryset

jonathan-s:
Only allowing published content to be available to be queried in the monkeypatched manager breaks any time `get` is used for draft content. 

~~It is more intuitive if all content is queryable as normal in the monkeypatched modelmanager, this won't also break any code that tries to access versionend content that is other than `published`.~~

~~Instead, it would be better to have published queryset on the manager. (not in this PR yet).~~

An alternative way of solving this would then be to allow all states to be accessible when making a get query. This would allow the use case of when looking up an entry through pk which would otherwise return `DoesNotExist`. 

**How to test manually**

Edit a page, add a new panel. Notice how when adding that new panel it won't reload the page as it did before.

### DIFF
--- a/djangocms_versioning/managers.py
+++ b/djangocms_versioning/managers.py
@@ -1,4 +1,29 @@
+from django.db import models
+from django.db.models.sql.where import WhereNode
+
 from .constants import PUBLISHED
+
+
+class PublishedQuerySet(models.QuerySet):
+
+    def get(self, *args, **kwargs):
+        """
+        A get query should always be able to access entries which are not in
+        a published state.
+        """
+        where_children = self.query.where.children
+        all_except_published = [
+            lookup for lookup in where_children
+            if not (
+                lookup.lookup_name == 'exact' and
+                lookup.rhs == 'published' and
+                lookup.lhs.field.name == 'state'
+            )
+        ]
+
+        self.query.where = WhereNode()
+        self.query.where.children = all_except_published
+        return super().get(*args, **kwargs)
 
 
 class PublishedContentManagerMixin:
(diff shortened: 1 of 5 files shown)


djangocms_versioning/managers.py
```diff
@@ -9,6 +9,4 @@ def get_queryset(self):
         """Limit query to published content
         """
         queryset = super().get_queryset()
-        if not self.versioning_enabled:
-            return queryset
-        return queryset.filter(versions__state=PUBLISHED)
+        return queryset
```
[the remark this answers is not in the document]
jonathan-s: It would be great if we could find out the reason for limiting the queryset in this way. From the code that I've read so far I haven't seen a good reason to keep it this way.

djangocms_versioning/managers.py
```diff
@@ -1,14 +1,39 @@
...
 
+class PublishedQuerySet(models.QuerySet):
+
+    def get(self, *args, **kwargs):
+        """
+        A get query should always be able to access entries which are not in
+        a published state.
+        """
+        where_children = self.query.where.children
+        all_except_published = [
```
Aiky30: @jonathan-s What is the benefit of this approach over an annotated query?

Aiky30: I would have thought that the query would need to be processed, returned and then looped through via Python before potentially being filtered further requiring another trip to the DB as any query on Page or Filer would hammer the .get method.

jonathan-s: From my understanding when you make an annotated query (ie `.annotate(...)`) you're adding extra data for the `SELECT` part of the query. Since the query is filtered in the `WHERE` part (ie `WHERE "djangocms_versioning_version"."state" = 'published'`) this is what we want to get rid of. Using annotate won't solve that. 

> I would have thought that the query would need to be processed, returned and then looped through via Python before potentially being filtered further requiring another trip to the DB as any query on Page or Filer would hammer the .get method.

Not sure what exactly you're after here. Do you think you could clarify if the above doesn't already answer your question.

Aiky30:
Versioning was always designed to never return a draft or archived version to ensure that nothing could be seen by a visitor by accident. It's a fail safe that should never be undone.

PR #135, opened 2018-10-24
Refactor compare view

chaosk:
Move both object ids to query params

PR #192, opened 2019-10-09
Refactor compare view

jonathan-s:
I've checked out the refactor by @chaosk and it checks out. Approved by me. This PR closes #135

### DIFF
--- a/djangocms_versioning/admin.py
+++ b/djangocms_versioning/admin.py
@@ -12,7 +12,9 @@ from django.template.loader import render_to_string, select_template
 from django.template.response import TemplateResponse
 from django.urls import reverse
 from django.utils.encoding import force_text
+from django.utils.formats import localize
 from django.utils.html import format_html, format_html_join
+from django.utils.timezone import localtime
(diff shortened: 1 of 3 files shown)


Aiky30 (changes requested):
After testing this change it is apparent that it is not a fully working example, it's also missing Frontend integration. The best solution is to start again from the original PR: https://github.com/divio/djangocms-versioning/pull/135 and add the new left and right query parameters in the js: 
- https://github.com/divio/djangocms-versioning/blob/aa7a16fe275a6d8a41781ffb1e10427de917c2d5/djangocms_versioning/static/djangocms_versioning/js/base.js#L14
- https://github.com/divio/djangocms-versioning/blob/aa7a16fe275a6d8a41781ffb1e10427de917c2d5/djangocms_versioning/static/djangocms_versioning/js/base.js#L42

jonathan-s:
@Aiky30 Unfortunately since I don't really use djangocms-versioning personally I won't spend my free time on making these changes.

PR #197, opened 2019-12-09
comment out 0012 data migration forward logic

joshyu:
Remove 0012 data migration script because versions will always be created after this point, and it is causing dependency problems with other packages tests. Anyone that requires this migration can install version 0.0.23, which is the last version with it enabled.

### DIFF
--- a/djangocms_versioning/migrations/0012_create_version_numbers.py
+++ b/djangocms_versioning/migrations/0012_create_version_numbers.py
@@ -18,29 +18,32 @@ def _distinct_groupers(versionable):
 def forwards(app_registry, schema_editor):
     """
     Assign version numbers to existing data
+
+    Removed because versions will always be created after this point, and it is causing dependency problems with other packages tests.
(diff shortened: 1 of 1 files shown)


djangocms_versioning/migrations/0012_create_version_numbers.py
```diff
@@ -18,29 +18,32 @@ def _distinct_groupers(versionable):
...
-            version_list = Version.objects.filter(
-                object_id__in=content_objects, content_type_id=content_type.pk
-            ).order_by("pk")
-
-            previous_number = 0
-            for version in version_list:
-                version.number = previous_number + 1
-                previous_number = version.number
-                version.save()
+    # versioning_extension = apps.get_app_config("djangocms_versioning").cms_extension
```
Aiky30: Hi @joshyu can you please add a return here:

```
def forwards(app_registry, schema_editor):
    """
    ...
    """
    return 



```

PR #199, opened 2020-01-09
Support django2.2

jonathan-s:
Currently it's failing so this a WIP investigation to figure out why. 

TypeError: Cannot use cached_property instance without calling `__set_name__()` on it.

when `__set_name__` is being called on the cached property, the cached property function is being replaced by the real function. 

When `__set_name__` is not called, this replacement does not happen, which executes the cached property function which subsequently throws an error. 

So for some reason cached property's `__set_name__` is not being called even though it should be. 

Django had a similar bug where this happened: 
- https://github.com/django/django/pull/10763/files
- https://code.djangoproject.com/ticket/29970

This is the error that is being thrown for `cached_property`
https://github.com/django/django/blob/stable/2.2.x/django/utils/functional.py#L31

Docs for cached_property in django 2.2: https://docs.djangoproject.com/en/2.2/ref/utils/#django.utils.functional.cached_property

**edit** 
I'm now 80% sure it's because we are monkey patching content_renderer and structure_renderer, right here: https://github.com/divio/djangocms-versioning/blob/master/djangocms_versioning/monkeypatch/toolbar.py

Changing `cached_property` to a `property` confirms that fact. It is the issue. A different issue crops up if you make that change.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -43,7 +43,7 @@ setup(
     zip_safe=False,
     tests_require=TEST_REQUIREMENTS,
     dependency_links=[
-        "https://github.com/jonathan-s/django-cms/tarball/django-2.2#egg=django-cms-4.0.0",
+        "http://github.com/divio/django-cms/tarball/release/4.0.x#egg=django-cms-4.0.0",
         "https://github.com/divio/djangocms-text-ckeditor/tarball/support/4.0.x#egg=djangocms-text-ckeditor-4.0.x",
     ]
 )
--- a/setup.py
+++ b/setup.py
@@ -4,7 +4,7 @@ import djangocms_versioning
 
 
 INSTALL_REQUIREMENTS = [
-    "Django>=1.11,<2.2",
+    "Django>=1.11,<3.0",
     "django-cms",
     "django-fsm>=2.6,<2.7"
 ]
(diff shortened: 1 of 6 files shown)


setup.py
```diff
@@ -43,7 +43,7 @@
     zip_safe=False,
     tests_require=TEST_REQUIREMENTS,
     dependency_links=[
-        "http://github.com/divio/django-cms/tarball/release/4.0.x#egg=django-cms-4.0.0",
+        "https://github.com/jonathan-s/django-cms/tarball/django-2.2#egg=django-cms-4.0.0",
```
jonathan-s: Needs to be replaced with the following once again when https://github.com/divio/djangocms-versioning/pull/199 is merged. 

```suggestion
        "http://github.com/divio/django-cms/tarball/release/4.0.x#egg=django-cms-4.0.0",
```

tox.ini
```diff
@@ -14,8 +14,9 @@ deps =
     dj111: Django>=1.11,<2.0
     dj20: Django>=2.0,<2.1
     dj21: Django>=2.1,<2.2
+    dj22: Django>=2.2,<3.0
 
-    cms40: https://github.com/divio/django-cms/archive/release/4.0.x.zip
+    cms40: https://github.com/jonathan-s/django-cms/archive/django-2.2.zip
```
jonathan-s: Needs to be replaced with the following once the patch for 2.2 is in place for django-cms. 

```suggestion
    cms40: https://github.com/divio/django-cms/archive/release/4.0.x.zip
```

PR #201, opened 2020-01-14
Be able fetch a page in extensions even if it is draft

jonathan-s:
This allows you to be able to copy the page with pagecontent.

### DIFF
--- a/djangocms_versioning/monkeypatch/extensions.py
+++ b/djangocms_versioning/monkeypatch/extensions.py
@@ -1,0 +1,21 @@
+from cms.extensions.extension_pool import ExtensionPool
+from cms.models.titlemodels import PageContent
+
+
+def _copy_title_extensions(self, source_page, target_page, language, clone=False):
+    source_title = source_page.pagecontent_set.get(language=language)
+    if target_page:
+        # the line below has been modified to accomodate versioning.
+        target_title = PageContent._original_manager.filter(
+            page=target_page, language=language
+        )
+    else:
+        target_title = source_title.publisher_public
+    for extension in self.title_extensions:
+        for instance in extension.objects.filter(extended_object=source_title):
+            if clone:
+                instance.copy(target_title, language)
+            else:
+                instance.copy_to_public(target_title, language)
+
+ExtensionPool._copy_title_extensions = _copy_title_extensions
--- a/djangocms_versioning/monkeypatch/extensions.py
+++ b/djangocms_versioning/monkeypatch/extensions.py
@@ -3,8 +3,14 @@ from cms.models.titlemodels import PageContent
 
 
 def _copy_title_extensions(self, source_page, target_page, language, clone=False):
-    # the line below has been modified to accomodate versioning.
-    # import pdb; pdb.set_trace()
+    """
+    djangocms-cms/extensions/admin.py, last changed in: divio/django-cms@2894ae8
+
+    The existing method ExtensionPool._copy_title_extensions will only ever
+    get published versions, we change the queries to get the latest draft version
+    with the _original_manager
+    """
+
     source_title = PageContent._original_manager.filter(
         page=source_page, language=language
     ).first()
--- a/djangocms_versioning/monkeypatch/extensions.py
+++ b/djangocms_versioning/monkeypatch/extensions.py
@@ -3,12 +3,16 @@ from cms.models.titlemodels import PageContent
 
 
 def _copy_title_extensions(self, source_page, target_page, language, clone=False):
-    source_title = source_page.pagecontent_set.get(language=language)
+    # the line below has been modified to accomodate versioning.
+    # import pdb; pdb.set_trace()
+    source_title = PageContent._original_manager.filter(
+        page=source_page, language=language
+    ).first()
     if target_page:
         # the line below has been modified to accomodate versioning.
         target_title = PageContent._original_manager.filter(
             page=target_page, language=language
-        )
+        ).first()
     else:
         target_title = source_title.publisher_public
     for extension in self.title_extensions:
--- a/djangocms_versioning/monkeypatch/extensions.py
+++ b/djangocms_versioning/monkeypatch/extensions.py
@@ -22,4 +22,5 @@ def _copy_title_extensions(self, source_page, target_page, language, clone=False
             else:
                 instance.copy_to_public(target_title, language)
 
+
 ExtensionPool._copy_title_extensions = _copy_title_extensions
--- a/djangocms_versioning/test_utils/factories.py
+++ b/djangocms_versioning/test_utils/factories.py
@@ -169,8 +169,8 @@ class PageContentFactory(factory.django.DjangoModelFactory):
     changed_by = FuzzyText(length=12)
     in_navigation = FuzzyChoice([True, False])
     soft_root = FuzzyChoice([True, False])
-    template = FuzzyText(length=12)
     limit_visibility_in_menu = constants.VISIBILITY_USERS
+    template = 'page.html'
     xframe_options = FuzzyInteger(0, 25)
 
     class Meta:
--- a/tests/test_monkeypatch.py
+++ b/tests/test_monkeypatch.py
@@ -50,7 +50,7 @@ class MonkeypatchExtensionTestCase(CMSTestCase):
         self.new_page.title_cache[pagecontent.language] = new_page_content
 
     def test_copy_extensions(self):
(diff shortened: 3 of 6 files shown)


djangocms_versioning/monkeypatch/extensions.py
```diff
@@ -0,0 +1,26 @@
+from cms.extensions.extension_pool import ExtensionPool
+from cms.models.titlemodels import PageContent
+
+
+def _copy_title_extensions(self, source_page, target_page, language, clone=False):
+    # the line below has been modified to accomodate versioning.
```
Aiky30: Can we add a comment as either a docstring or above the function explaining why we have made changes here.

Something like:  The existing method ExtensionPool._copy_title_extensions will only ever get published versions, we change the queries to get the latest draft version.

djangocms_versioning/monkeypatch/extensions.py
```diff
@@ -0,0 +1,26 @@
...
+    # import pdb; pdb.set_trace()
+    source_title = PageContent._original_manager.filter(
```
Aiky30: Should we not be using _base_manager vs _original_manager?

djangocms_versioning/monkeypatch/extensions.py
```diff
@@ -0,0 +1,26 @@
...
+        page=source_page, language=language
+    ).first()
+    if target_page:
+        # the line below has been modified to accomodate versioning.
+        target_title = PageContent._original_manager.filter(
```
Aiky30: Looking at the original commit is there a clever way to change "pagecontent_set"? If there was out monkey patch would literally just change the set returned. Food for thought.

djangocms_versioning/monkeypatch/extensions.py
```diff
@@ -0,0 +1,26 @@
+from cms.extensions.extension_pool import ExtensionPool
+from cms.models.titlemodels import PageContent
+
+
+def _copy_title_extensions(self, source_page, target_page, language, clone=False):
```
Aiky30: One thing thta we have never done which is bad is add a comment referring to the commit / code state that the code is copied from. That would make upgrading it, or even changing it in the future much easier. you could guess from the date of the PR but being explicit is always better. Something like: 

# Taken from djangocms-cms/extensions/admin.py, last changed in: https://github.com/divio/django-cms/commit/2894ae8bcf92092d947a097499c01ab2bbb0e6df

Strangely my commit.

tests/test_monkeypatch.py
```diff
@@ -1,17 +1,64 @@
...
+        site = Site.objects.first()
+        self.new_page = self.page.copy(
+            site=site,
+            parent_node=self.page.node.parent,
+            translations=False,
+            permissions=False,
+            extensions=False,
+        )
+
+        new_page_content = api.create_title(
```
Aiky30: Strange to use factories and then the api for creating a new PageContent item. Why is this?

tests/test_monkeypatch.py
```diff
@@ -1,17 +1,64 @@
...
+        self.new_page.title_cache[pagecontent.language] = new_page_content
+
+    def test_copy_extensions(self):
+        '''Try to copy the extension, without the monkeypatch this tests fails'''
+        extension_pool = ExtensionPool()
+        extension_pool.page_extensions = set([TestPageExtension])
+        extension_pool.title_extensions = set([TestTitleExtension])
+        extension_pool.copy_extensions(
+            self.page, self.new_page, languages=['de']
+        )
```
Aiky30: No asserts, is this even testing anything? Would it be a pass? I'm not actually sure what would happen, never tried.

tests/test_monkeypatch.py
```diff
@@ -1,17 +1,64 @@
...
+            slug='slug',
+            path='/path/here',
+            title=pagecontent.title,
+            template=pagecontent.template,
+            created_by=user
+        )
+        self.new_page.title_cache[pagecontent.language] = new_page_content
+
+    def test_copy_extensions(self):
+        '''Try to copy the extension, without the monkeypatch this tests fails'''
```
Aiky30: try not to add different styles of docstrings to keep the code consistent. It's an art to shape shift your own style to keep codebases consistent.

djangocms_versioning/monkeypatch/extensions.py
jonathan-s: Does not work for _base_manager. We need the _original_manager.

tests/test_monkeypatch.py
jonathan-s: Did not work properly with factories, and using api.create_title is in itself a kind of factory, no? :)

tests/test_monkeypatch.py
jonathan-s: The bane of my existence. ''' forever in my heart <3 (I will change this).

tests/test_monkeypatch.py
jonathan-s: It does test something. It tests that there are no failures. Removing the monkeypatch and copying the extension fails. There is no self.assertNotRaises, I did add a comment that says that without the monkeypatch it fails.

djangocms_versioning/monkeypatch/extensions.py
jonathan-s: I will leave this as it is for now.

tests/test_monkeypatch.py
Aiky30: Create title creates the models themselves unlike a factory which only creates what you specify for it to create.

tests/test_monkeypatch.py
Aiky30: Ok. It should probably explain why it would fail though because if it does start to fail anyone fixing / rewoirkign the test will need to know why the test and the monkey patch are there.

djangocms_versioning/test_utils/factories.py
```diff
@@ -158,7 +158,7 @@ class PageContentFactory(factory.django.DjangoModelFactory):
     changed_by = FuzzyText(length=12)
     in_navigation = FuzzyChoice([True, False])
     soft_root = FuzzyChoice([True, False])
-    template = FuzzyText(length=12)
+    template = 'page.html'
```
Aiky30: Why was the template changed?

djangocms_versioning/monkeypatch/extensions.py
Aiky30: Ok. I don't have a solution for that so fine to leave.

djangocms_versioning/test_utils/factories.py
jonathan-s: Templates was changed as it must exist to be able to use it. `FuzzyText` templates does not exist.

tests/test_monkeypatch.py
jonathan-s: Will fix this.

jonathan-s:
```suggestion
        )
        # No asserts, this test originally failed because the versioned manager was called
        # in copy_extensions, now we call the original manager instead
        # https://github.com/divio/djangocms-versioning/pull/201/files#diff-fc33dd7b5aa9b1645545cf48dfc9b4ecR19 
```

jonathan-s: @Aiky30 Would this comment do?

Aiky30: @jonathan-s I can't commit your suggestion as I don't have write access to your branch.

djangocms_versioning/test_utils/factories.py
Aiky30: Thinking about this, this should still be randomly generated. If you want to use the page then you set that parameter when you use it. Most tests don't require it so it's right that it's just a random value. The internal codebase does this a little better where it selects from a list of templates.

tests/test_monkeypatch.py
Aiky30: I think as you are testing the actual functionality of what happens with the slug and path on creation and changes it is ok to use the actual models vs the factories. 

We do have a PageUrl factory for tests that need to access a page.

djangocms_versioning/test_utils/factories.py
jonathan-s: Factories should have sane defaults. FuzzyText is not a sane default considering that a correct value is a template that exists. 

I would agree that a randomly picked template that exist is better than a single value that exist. 

Generating text that has no meaning, when in fact the underlying value should point to something else is not something that is desirable. So I disagree with you on this point.

jonathan-s:
🎉

PR #202, opened 2020-01-15
Deleting the last version that exist in a grouper deletes the grouper as well

jonathan-s:
When you delete the last version that exist in a grouper it will delete the grouper itself.

### DIFF
--- a/djangocms_versioning/admin.py
+++ b/djangocms_versioning/admin.py
@@ -26,7 +26,7 @@ from .compat import DJANGO_GTE_21
 from .constants import ARCHIVED, DRAFT, PUBLISHED, UNPUBLISHED
 from .exceptions import ConditionFailed
 from .forms import grouper_form_factory
-from .helpers import get_editable_url, get_preview_url, version_list_url
+from .helpers import get_admin_url, get_editable_url, get_preview_url, version_list_url
 from .models import Version
 from .versionables import _cms_extension
 
@@ -709,7 +709,6 @@ class VersionAdmin(admin.ModelAdmin):
     def discard_view(self, request, object_id):
         """Discards the specified version"""
         version = self.get_object(request, unquote(object_id))
-
         if version is None:
             raise Http404
 
@@ -737,13 +736,16 @@ class VersionAdmin(admin.ModelAdmin):
             return render(
                 request, "djangocms_versioning/admin/discard_confirmation.html", context
             )
-        else:
-            version_url = version_list_url(version.content)
-            if request.POST.get("discard"):
-                version.delete()
 
-            # Redirect
-            return redirect(version_url)
+        version_url = version_list_url(version.content)
+        if request.POST.get("discard"):
+            ModelClass = version.content.__class__
+            deleted = version.delete()
+            if deleted[1]['last']:
+                # need a success message as well.
+                version_url = get_admin_url(ModelClass, 'changelist')
+
+        return redirect(version_url)
 
     def compare_view(self, request, object_id):
         """Compares two versions
--- a/djangocms_versioning/admin.py
+++ b/djangocms_versioning/admin.py
@@ -816,6 +816,15 @@ class VersionAdmin(admin.ModelAdmin):
             )
         except (TypeError, ValueError):
             grouper = None
+        else:
+            if grouper is None:
+                # no exception and no grouper, thus the querydict is invalid
+                url = reverse(
+                    "admin:{app}_{model}_grouper".format(
+                        app=self.model._meta.app_label, model=self.model._meta.model_name
+                    ),
+                )
+                return redirect(url)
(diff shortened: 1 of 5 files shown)


djangocms_versioning/admin.py
```diff
@@ -816,6 +815,15 @@ def changelist_view(self, request, extra_context=None):
             )
         except (TypeError, ValueError):
             grouper = None
+        else:
+            if grouper is None:
```
Aiky30: Is this check redundant?

djangocms_versioning/admin.py
```diff
@@ -816,6 +815,15 @@ def changelist_view(self, request, extra_context=None):
             )
         except (TypeError, ValueError):
```
Aiky30: Do we know in what circumstances the following are caught: TypeError, ValueError? Comments would have helped us here

djangocms_versioning/models.py
```diff
@@ -95,6 +95,19 @@ class Meta:
     def __str__(self):
         return "Version #{}".format(self.pk)
 
+    def delete(self, using=None, keep_parents=False):
```
Aiky30: Can you add a comment or docstring to explain why we are deleting here. We are seriously lacking comments in the code.

djangocms_versioning/models.py
```diff
@@ -95,6 +95,19 @@ class Meta:
...
+        ContentModel = self.content._meta.model
+        grouper_name = self.grouper._meta.model_name
+        querydict = {'{}__pk'.format(grouper_name): self.grouper.pk}
+        count = ContentModel._original_manager.filter(**querydict).count()
+
+        grouper = self.grouper
```
Aiky30: Is the assignment here redundant when it's only used once: grouper.delete()

tests/test_admin.py
```diff
@@ -1985,6 +1985,22 @@ def test_missing_grouper(self):
             response, "/en/admin/djangocms_versioning/pollcontentversion/?e=1"
         )
 
+    def test_missing_grouper_does_not_exist(self):
+        """Go to changelist with a grouper that does not exist in querystring
```
Aiky30: Was this functionality of redirecting specified by Kuldeep?

djangocms_versioning/admin.py
jonathan-s: It's basically because of `int()` and `.GET.get()`

djangocms_versioning/admin.py
jonathan-s: No. if the grouper is None, not because there was an error, but because no grouper exists we want to redirect to the grouper page where the user can select a valid grouper. 

If the there is an exception an error page will be shown instead.

tests/test_admin.py
jonathan-s: It was not, but from a UX stand point it made sense. I know, I know I've been wrong before.

djangocms_versioning/models.py
jonathan-s: It is not redundant, when we execute the `delete` function `self` no longer contains grouper. Test fails if you change this, so there is a safety net for this.

djangocms_versioning/admin.py
jonathan-s: Do you want a comment for this here?

jonathan-s:
@Aiky30 I there are a few more changes that need to get in based on the discussion that we had on tuesday. Do you need this finished by friday?

djangocms_versioning/admin.py
```diff
@@ -737,14 +736,17 @@ def discard_view(self, request, object_id):
...
-                version.delete()
 
-            # Redirect
-            return redirect(version_url)
+        version_url = version_list_url(version.content)
+        if request.POST.get("discard"):
+            ModelClass = version.content.__class__
+            deleted = version.delete()
+            if deleted[1]['last']:
+                # need a success message as well.
```
jonathan-s:
```suggestion

```

jonathan-s: This should be deleted.

jonathan-s: @Aiky30 You can commit this change. Apparently only users with write permission can commit things from github.

djangocms_versioning/admin.py
```diff
@@ -737,14 +736,17 @@ def discard_view(self, request, object_id):
...
+                version_url = get_admin_url(ModelClass, 'changelist')
+                self.message_user(request, 'The last version has been deleted')
```
Aiky30: This string will need to be translatable with gettext

djangocms_versioning/admin.py
Aiky30: I also cannot commit this suggestion because I don't have the permission on your repo.

PR #203, opened 2020-01-21
Urls no longer need to be updated

jonathan-s:
The urls does not need to be updated at all. The reason urls
were presumably updated in the past was to avoid that you could
visit a page if the page was in draft mode.

This has been fixed in django-cms. A page that is in draft
will return an EmptyPageContent, if it is empty it won't render.

See divio/django-cms#6803

Added tests are mostly for documentation purposes to see expected behaviour. 

From @jonathan-s 11th Feb 2020 justifying why we need to make this change:

On the off chance that there is a page that uses the old structure and there is a change that requires it I've left it in. 
 
So the scenario is the following. User creates a page, pagecontent is in draft. User publishes pagecontent. They change the path to something custom (this sets managed to False in that url). Prior to this patch everything works fine. With this setup.
 
User creates a new version. User publishes this. This means that the prior version gets unpublished. So the page url would have been set to none. New version gets published. Since only managed urls gets updated (there is no way to refresh unmanaged urls as they are static and the static thing was just deleted). 
 
With this patch we avoid setting urls to None since this is no longer necessary. We've solved the findability of a page in core-cms. 

@Aiky30 would like to add that as there were no tests for the changes originally it's very difficult to know why the features were there in the first place, we could only assume!!

Changes made with no reasoning of why:
- https://github.com/divio/djangocms-versioning/pull/61
- https://github.com/divio/djangocms-versioning/pull/62

### DIFF
--- a/tests/test_cms_config.py
+++ b/tests/test_cms_config.py
@@ -3,18 +3,121 @@ from unittest.mock import Mock, patch
 
 from django.apps import apps
 from django.contrib import admin
+from django.contrib.sites.models import Site
 from django.core.exceptions import ImproperlyConfigured
+from django.test import RequestFactory
+from django.utils.text import slugify
 
+from cms.admin.forms import ChangePageForm
+from cms.models import Page
 from cms.test_utils.testcases import CMSTestCase
 
 from djangocms_versioning.admin import VersionAdmin, VersioningAdminMixin
 from djangocms_versioning.cms_config import VersioningCMSConfig, VersioningCMSExtension
+from djangocms_versioning.constants import DRAFT
 from djangocms_versioning.datastructures import VersionableItem, default_copy
 from djangocms_versioning.models import Version
+from djangocms_versioning.test_utils import factories
 from djangocms_versioning.test_utils.blogpost.cms_config import BlogpostCMSConfig
 from djangocms_versioning.test_utils.blogpost.models import BlogContent, Comment
 from djangocms_versioning.test_utils.polls.cms_config import PollsCMSConfig
 from djangocms_versioning.test_utils.polls.models import Poll, PollContent
+
+
+req_factory = RequestFactory()
+
+
+class PageContentVersioningBehaviourTestCase(CMSTestCase):
+
+    def setUp(self):
+        self.site = Site.objects.get_current()
+        self.user = self.get_superuser()
+        self.language = 'en'
+        self.title = 'test page'
+
+        self.version = factories.PageVersionFactory(content__language='en', state=DRAFT,)
+        factories.PageUrlFactory(
+            page=self.version.content.page,
+            language='en',
+            path=slugify(self.title),
+            slug=slugify(self.title),
+        )
+
+        self.page = self.version.content.page
+        self.content = self.version.content
+
+    def test_saving_draft_has_pageurl(self):
+        """A draft has pageurl as normal"""
+        url = self.page.get_urls().first()
+
+        self.assertEqual(self.version.state, DRAFT)
+        self.assertEqual(url.path, slugify(self.title))
+
+    def test_publishing_draft_retains_pageurl(self):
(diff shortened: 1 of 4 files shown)


tests/test_cms_config.py
```diff
@@ -17,6 +24,96 @@
...
+        url.save()
+
+        self.version.publish(self.user)
+        v2 = self.version.copy(self.user)
+        v2.publish(self.user)
+
+        page = Page.objects.get(pk=self.page.pk)
+        url = page.get_urls().first()
+
+        self.assertEqual(url.path, slugify(self.title))
```
jonathan-s: This test is the one that will fail if we change add https://github.com/divio/djangocms-versioning/pull/203/files#diff-8b290a1c3759be311fb10a9b1a4faa88L216

tests/test_cms_config.py
```diff
@@ -17,6 +24,96 @@
...
+
+    def test_published_version_with_new_version_retains_pageurl_unmanaged(self):
+        '''We have a published, we create a new version and publish that.
+        The url in question is unmanaged and stays the same.'''
```
Aiky30: Can we add here that: the path was previously set to None which deleted the contents when a content editor had added an overwrite url manually which is store in the path field.

tests/test_cms_config.py
```diff
@@ -17,6 +24,96 @@
...
+
+class PageContentVersioningBehaviourTestCase(CMSTestCase):
+
+    def setUp(self):
+        self.site = Site.objects.get_current()
+        self.user = self.get_superuser()
+        self.language = 'en'
+        self.title = 'test page'
+
+        self.page = create_page(
```
Aiky30: We shouldn't be using the api to create a page. The PageContent and a PageUrl entry should be all that we need:

version = PageContentVersionFactory(content__language='en', state=PUBLISHED,)
PageUrlFactory(page=version.content.page, language='en', path="hello", slug="hello",)

tests/test_cms_config.py
```diff
@@ -17,6 +24,96 @@
...
+            in_navigation=True,
+            limit_visibility_in_menu=None,
+            site=self.site,
+            created_by=self.user,
+        )
+        self.content = PageContent._original_manager.get(page=self.page.pk)
+        self.version = self.content.versions.all()[0]
+
+    def test_saving_draft_has_pageurl(self):
+        '''A draft has pageurl as normal'''
```
Aiky30: """

tests/test_cms_config.py
```diff
@@ -17,6 +24,96 @@
...
+
+    def test_publishing_draft_retains_pageurl(self):
+        '''We publish a version, pageurl is still the same'''
+        self.version.publish(self.user)
+
+        url = self.page.get_urls().first()
+        self.assertEqual(url.path, slugify(self.title))
+
+    def test_published_version_with_new_version_retains_pageurl(self):
+        '''We have a published, we create a new version and publish that.
```
Aiky30: Slight grammar change, I'm not sure what is meant by "We have a published," Do you mean we have a published page, or we have published, ....

tests/test_cms_config.py
```diff
@@ -17,6 +24,96 @@
...
+        v2 = self.version.copy(self.user)
+        v2.publish(self.user)
+
+        page = Page.objects.get(pk=self.page.pk)
+        url = page.get_urls().first()
+
+        self.assertEqual(url.path, slugify(self.title))
+
+    def test_published_version_with_new_version_retains_pageurl_unmanaged(self):
+        '''We have a published, we create a new version and publish that.
```
Aiky30: Slight grammar change, I'm not sure what is meant by "We have a published,"

tests/test_cms_config.py
jonathan-s: Why shouldn't we be using the api to create a page?

PR #206, opened 2020-02-24
Monkeypatch so that draft object can be retrieved

jonathan-s:
`RequestToolbarForm` in django-cms needs to be able to fetch objects that are in draft as well. A method `get_admin_model_object_by_id` was created in the following PR (https://github.com/divio/django-cms/pull/6806) to allow versioning to monkeypatch this method.


jonathan-s:
🎉

PR #207, opened 2020-02-24
Don't assume the grouper name, check the actual field name

jonathan-s:
We recently modified the delete method in this PR (https://github.com/divio/djangocms-versioning/pull/202) in that PR I made the somewhat unfortunate decision to assume the field name of the grouper. (This works fine as long as you don't explicitly name the grouper field name). 

This PR amends that mistake.

### DIFF
--- a/djangocms_versioning/models.py
+++ b/djangocms_versioning/models.py
@@ -99,12 +99,18 @@ class Version(models.Model):
         """Deleting a version deletes the grouper
         as well if we are deleting the last version."""
 
-        ContentModel = self.content._meta.model
-        grouper_name = self.grouper._meta.model_name
-        querydict = {'{}__pk'.format(grouper_name): self.grouper.pk}
-        count = ContentModel._original_manager.filter(**querydict).count()
+        def get_grouper_name(ContentModel, GrouperModel):
+            for field in ContentModel._meta.fields:
+                if getattr(field, 'related_model', None) == GrouperModel:
+                    return field.name
 
         grouper = self.grouper
+        ContentModel = self.content._meta.model
+
+        grouper_name = get_grouper_name(ContentModel, grouper._meta.model)
+        querydict = {'{}__pk'.format(grouper_name): grouper.pk}
+        count = ContentModel._original_manager.filter(**querydict).count()
+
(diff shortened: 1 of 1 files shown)


djangocms_versioning/models.py
```diff
@@ -99,12 +99,19 @@ def delete(self, using=None, keep_parents=False):
...
+            GrouperModel = grouper._meta.model
+            for field in ContentModel._meta.fields:
+                if getattr(field, 'related_model', None) == GrouperModel:
+                    return field.name
+
+        grouper = self.grouper
         ContentModel = self.content._meta.model
-        grouper_name = self.grouper._meta.model_name
+
+        grouper_name = get_grouper_name(ContentModel, grouper)
```
Aiky30: minor thing, why create grouper when it's just a reference to self.grouper? Compounded by having self.grouper.pk on the line below too.

djangocms_versioning/models.py
```diff
@@ -99,12 +99,19 @@ def delete(self, using=None, keep_parents=False):
         """Deleting a version deletes the grouper
         as well if we are deleting the last version."""
 
+        def get_grouper_name(ContentModel, grouper):
```
Aiky30: Wondering if this should be a helper in helpers.py, get_grouper_name_from_content_model(). I guess this depends on whether it could be used again in this package or by others.

jonathan-s:
You asked exactly the same thing the last time. :)

grouper is not a reference. It is its own object and we need it once we've
done the deletion.

I haven't seen the need for this elsewhere so it seems like this would be
premature optimization. I made it its own method because it makes things
more readable.

djangocms_versioning/models.py
jonathan-s: You asked exactly the same thing the last time. :)

grouper is not a reference. It is its own object and we need it once we've
done the deletion.

djangocms_versioning/models.py
jonathan-s: I haven't seen the need for this elsewhere so it seems like this would be
premature optimization. I made it its own method because it makes things
more readable.

PR #211, opened 2020-03-10
Mock filter_admin_model

jonathan-s:
By mocking this method we will be able to filter from the entire queryset
rather than the queryset limited to published objects

The reason for introducing this in the first place was that the toolbar needs to access to be able to access an object that is draft and it should not rely on `page.get_title` to fetch the correct object as that method will fetch fallbacks which may not be correct for our purposes. 

This PR depends on: https://github.com/divio/django-cms/pull/6817 for tests to succeed.


jonathan-s:
Closing as the PR mentioned above has also been closed.

PR #214, opened 2020-03-13
Visualize and fix issues with rendering toolbar and pagecontent

jonathan-s:
This PR adds several tests regarding rendering of pagecontent. 

In the PR https://github.com/divio/django-cms/pull/6820 we add hooks to get_admin_title_obj that we can re-use here. 

One of the issues is that the pagecontent (title_obj) can't be accesssed when you are in draft. Instead it's accessing the fallback. Which is not desirable. 

This leads to not being able to edit a pagecontent in another language. The language also doesn't link to the correct urls because of this. 

The tests check so that when you select a given language you are indeed viewing the correct content. And if it is a draft an unauthenticated user won't be able to view the content. 

In short this PR fixes issues that appear with the VersionContentRenderer > https://github.com/divio/djangocms-versioning/blob/master/djangocms_versioning/plugin_rendering.py#L45-L47

VersionContentRenderer avoids rendering a page entirely. So it avoids that the following method is triggering at all.  > https://github.com/divio/django-cms/blob/release/4.0.x/cms/plugin_rendering.py#L352 

If we were to adapt this for versions in some way. We would have to use pagecontent rather than page.


tests/test_views.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def test_normal_view_logged_out(self):
+        """A page in draft should give you a 404 if you visit the public url"""
+        self.client.logout()
+        response = self.client.get(self.german_url)
+
+        self.assertEqual(response.status_code, 404)
+
+    def test_absolute_url_of_draft(self):
+        self.assertEqual(self.german.get_absolute_url(), '/de/german-content/')
```
jonathan-s: The only test that is failing after the PR in django-cms is failing. So this should probably be fixed as well.

jonathan-s:
Speaking with Kuldeep I have also confirmed that visiting the real page url in draft should generate a 404. 

This means that the language buttons would need to redirect to the edit url instead of the detail view url.

jonathan-s:
Superseeded by #218 and #217.

PR #216, opened 2020-03-20
Move duplicate view from pageadmin to versioning

jonathan-s:
As described in this PR here -> https://github.com/FidelityInternational/djangocms-pageadmin/pull/45

We want to move the duplicate view from pageadmin to versioning. The reason for the need to make this move of the code is so that a person won't be required to install djangocms-pageadmin to not get an error when duplicating pagecontent. 

Next step is to remove code in pageadmin that is no longer needed.

### DIFF
--- a/tests/test_admin_using_versioning.py
+++ b/tests/test_admin_using_versioning.py
@@ -1,0 +1,173 @@
+from unittest.mock import patch
+
+import django
+from django.contrib import admin
+from django.contrib.sites.models import Site
+
+from cms.api import add_plugin
+from cms.models import PageContent, PageUrl
+from cms.test_utils.testcases import CMSTestCase
+from cms.utils.plugins import downcast_plugins
+
+from djangocms_versioning.test_utils.factories import (
+    PageContentWithVersionFactory,
+    PlaceholderFactory,
+)
+
+
+version = list(map(int, django.__version__.split('.')))
+GTE_DJ20 = version[0] >= 2
+
+
+class PageContentAdminTestCase(CMSTestCase):
+    def setUp(self):
+        self.modeladmin = admin.site._registry[PageContent]
+
+    def test_duplicate_url_is_replaced(self):
+        """
+        The old url /duplicate/ has been removed. But cms_pagecontent_duplicate
+        still exists.
+        """
+
+        urls = self.modeladmin.get_urls()
+        if GTE_DJ20:
+            duplicate_url = [u for u in urls if '/duplicate/' in u.pattern.regex.pattern]
+        else:
+            duplicate_url = [u for u in urls if '/duplicate/' in u.regex.pattern]
+        name_url = [u for u in urls if 'cms_pagecontent_duplicate' == u.name]
+
+        self.assertEqual(len(duplicate_url), 0)
+        self.assertEqual(len(name_url), 1)
+
+
+class DuplicateViewTestCase(CMSTestCase):
+    def test_obj_does_not_exist(self):
+        with self.login_user_context(self.get_superuser()), patch(
+            "django.contrib.messages.add_message"
+        ) as mock:
+            response = self.client.get(
+                self.get_admin_url(PageContent, "duplicate", "foo")
+            )
+
+        self.assertRedirects(response, "/en/admin/", target_status_code=302)
+        self.assertEqual(mock.call_count, 1)
+        self.assertEqual(mock.call_args[0][1], 30)  # warning level
+        self.assertEqual(
+            mock.call_args[0][2],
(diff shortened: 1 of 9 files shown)


tests/test_admin_using_versioning.py
```diff
@@ -0,0 +1,164 @@
...
+        with self.login_user_context(self.get_superuser()):
+            response = self.client.post(
+                self.get_admin_url(PageContent, "duplicate", pagecontent2.pk),
+                data={"site": Site.objects.first().pk, "slug": "bar"},
+                follow=True,
+            )
+        self.assertRedirects(response, self.get_admin_url(PageContent, "changelist"))
+        new_pagecontent = PageContent._base_manager.latest("pk")
+        new_placeholder = new_pagecontent.placeholders.get(slot="content")
+        self.assertEqual(PageContent._base_manager.count(), 3)
```
Aiky30: We should follow the AAA pattern with tests, this will need to be spaced out accordingly: https://jamescooke.info/arrange-act-assert-pattern-for-python-developers.html

djangocms_versioning/admin/using_versioning.py
```diff
@@ -0,0 +1,222 @@
+from django.conf.urls import url
```
Aiky30: I'm not so sure about the filename for this file, following where this is in the cms I would opt for somethign similar but generic, maybe admin/objectadmin.py or genericadmin.py

https://github.com/divio/django-cms/blob/release/4.0.x/cms/admin/pageadmin.py

djangocms_versioning/admin/using_versioning.py
```diff
@@ -0,0 +1,222 @@
...
+from django.urls import reverse
+from django.utils.translation import ugettext_lazy as _
+
+from cms import api
+from cms.extensions import extension_pool
+from cms.models import PageContent
+
+from .. import versionables
+from ..compat import DJANGO_GTE_21
+from ..forms import DuplicateForm
```
Aiky30: I don't like relative imports that scale over a a dir. Personal preference so not a big deal.

djangocms_versioning/admin/using_versioning.py
```diff
@@ -0,0 +1,222 @@
...
+        return super().get_readonly_fields(request, obj)
+
+    def has_change_permission(self, request, obj=None):
+        if obj and DJANGO_GTE_21:
+            version = Version.objects.get_for_content(obj)
+            return version.check_modify.as_bool(request.user)
+        return super().has_change_permission(request, obj)
+
+
+class VersioningPageContentAdminMixin(VersioningAdminMixin):
```
Aiky30: Part of the filename discussion, I'd see this as being in a file called admin/pageadmin.py as this is an exact comparison to the cms.

djangocms_versioning/templates/djangocms_versioning/admin/duplicate_confirmation.html
```diff
@@ -0,0 +1,29 @@
...
+    {{ block.super }}
+    {{ media }}
+    <script type="text/javascript" src="{% static 'admin/js/cancel.js' %}"></script>
+{% endblock %}
+
+{% block breadcrumbs %}{% endblock %}
+{% block bodyclass %}{{ block.super }} app-{{ opts.app_label }} model-{{ opts.model_name }} delete-confirmation{% endblock %}
+
+{% block content %}
+<p>{% trans "Are you sure you want to duplicate the following page?" %}</p>
```
Aiky30: These template files appear to be page specific, They should reflect that somehow either the folder they are in or the template filename

djangocms_versioning/admin/using_versioning.py
jonathan-s: Fixed

djangocms_versioning/templates/djangocms_versioning/admin/duplicate_confirmation.html
jonathan-s: Called it duplicate_page_confirmation

djangocms_versioning/admin/using_versioning.py
jonathan-s: Called pageadmin.py

tests/test_admin_using_versioning.py
jonathan-s: I've added carriage returns to separate it into blocks.

jonathan-s:
Andrew can choose to squash commits as he merges the PR :)

Aiky30:
This was a good effort to try and standardise a way of getting the packages to use a set of controls. The issue with this is that the ux and flow hasn't yet been agreed. When they are this PR can be referenced for guidelines to a possible solution.

PR #217, opened 2020-03-25
A more extensive test suite for testing views with versioning.

jonathan-s:
This testsuite is intended to test the behaviour of how authenticated and non-authenticated users view pages so that they can edit, create new translations, remove translations in an expected way.


Aiky30:
Seems to be incomplete WIP. Closing but will attempt to bring the tests that were created in when changes arise that affect the views.

PR #218, opened 2020-03-27
Separate rendering of toolbar and content

jonathan-s:
With this PR we can have one object for the toolbar while we have a different object for content. This manifests itself when we visit the live url. We should see the published page and in the toolbar we should be able to refer to the draft such as in the edit menu. 

- Tests can be found in #217
- Depends on https://github.com/divio/django-cms/pull/6827 to work properly.

### DIFF
--- a/djangocms_versioning/monkeypatch/page.py
+++ b/djangocms_versioning/monkeypatch/page.py
@@ -13,6 +13,40 @@ User = get_user_model()
 cms_extension = apps.get_app_config("cms").cms_extension
 
 
+def _get_admin_title_cache(page, language, fallback, force_reload):
+    from cms.utils import i18n
+    from django.utils.translation import get_language
+
+    def get_fallback_language(page, language):
+        fallback_langs = i18n.get_fallback_languages(language)
+        for lang in fallback_langs:
+            if page.admin_title_cache.get(lang):
+                return lang
+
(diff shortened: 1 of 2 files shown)


djangocms_versioning/monkeypatch/page.py
```diff
@@ -13,6 +13,40 @@
...
+
+    if not language:
+        language = get_language()
+
+    force_reload = (force_reload or language not in page.admin_title_cache)
+    if force_reload:
+        from cms.models import PageContent
+        titles = PageContent._original_manager.filter(page=page)
+        for title in titles:
+            page.admin_title_cache[title.language] = title
```
jonathan-s: One of the issues that could crop up here is that the cache is filled with a published pagecontent (so we won't check for new content), whereas what would have been more appropriate in a specific case is to give the draft pagecontent. The above depends on how long lived the cache is, if the cache only lives for a single request (which I think is the case) this shouldn't be an issue. Otherwise I'm not entirely sure how this should be handled. 

In the query above we could get several pagecontent with the same language (drafts and published) and we don't know which would be assigned to the cache in the end. So the query needs to be more specific.

Aiky30:
As this requires architectural changes it's better that they are made rather than hack around them here. Closing for now.

PR #219, opened 2020-03-29
fix prefetch error when current model has a m2m relation to grouper field

joshyu:
recently I met with a problem, when in some model, we contains a fields with manytomany relationship with the versioned model, then we will got a error when we want to filter the queryset with the filtered grouper list.

I fixed this case by adding a condition checker,  when `instance.field` return a instance of Model, then we will use the existing filter, if the return value is a instance of model.Manager, then It will need us to adjust the filter parameters.


djangocms_versioning/plugin_rendering.py
```diff
@@ -24,7 +26,13 @@ def prefetch_versioned_related_objects(instance, toolbar):
             qs = versionable.content_model.objects.all()
         related_field = getattr(instance, field.name)
         if related_field:
-            filters = {versionable.grouper_field_name: related_field}
+            if isinstance(related_field, models.Manager):
+                filters = {}
+                filter_key = f'{versionable.grouper_field_name}__in'
```
jonathan-s: f-strings is not supported by python 3.5. 
```suggestion
                filter_key = '{{}}__in'.format(versionable.grouper_field_name)
```

joshyu: I will change that.

joshyu:
@jonathan-s , @Aiky30 , @adam-murray 

I just added the test file to my changes. 
could you please help to review?

djangocms_versioning/plugin_rendering.py
```diff
@@ -24,14 +27,33 @@ def prefetch_versioned_related_objects(instance, toolbar):
...
             prefetch_cache = {versionable.grouper_field.remote_field.name: qs}
-            related_field._prefetched_objects_cache = prefetch_cache
+
+            # TODO refine it after understand prefetch in many2many field.
+            # because if `related_field` is ManyRelatedManager, it is temporary,
+            # we can't store `_prefetched_objectes_cache` to it, so I decide to store the
+            # prefetched value to model instance.
+            if is_related_manager:
+                try:
+                    instance._prefetched_objects_cache[field.name] = prefetch_cache
```
jonathan-s: If you use this instead I believe you'd eliminate the try ... except block entirely. 
```
getattr(instance, '_prefetch_objects_cache', {})[field.name] = prefetch_cache
```

Aiky30:
@joshyu What was the error that you saw to need this fix? Placeholders and plugins have many to many relationships, why don;t they error without this fix?

djangocms_versioning/plugin_rendering.py
```diff
@@ -24,14 +27,33 @@ def prefetch_versioned_related_objects(instance, toolbar):
             qs = versionable.content_model.objects.all()
         related_field = getattr(instance, field.name)
         if related_field:
-            filters = {versionable.grouper_field_name: related_field}
+            is_related_manager = isinstance(related_field, models.Manager)
+
+            if is_related_manager:
+                filter_key = '{}__in'.format(versionable.grouper_field_name)
+                filter_values = related_field.all()
```
Aiky30, line 34: I'm not sure this is the best way to prefetch a relationship with .all() because any filters added to the related values will then need to discard anything from this query adding overhead to this query and then having to call other queries anyway.

tests/test_monkeypatch.py
```diff
@@ -61,6 +63,56 @@ def test_content_renderer(self):
             CMSToolbar(request).content_renderer.__class__, VersionContentRenderer
         )
 
+    def test_prefetch_versioned_m2m_objects(self):
```
Aiky30, line 66: These tests are not testing the monkey patches, they are testing plugin rendering. Please move these test out of this file.

djangocms_versioning/plugin_rendering.py
```diff
@@ -24,14 +27,33 @@ def prefetch_versioned_related_objects(instance, toolbar):
...
+                filters = {versionable.grouper_field_name: related_field}
             # TODO Figure out grouping values-awareness
             # for extra fields other than hardcoded 'language'
             if "language" in versionable.extra_grouping_fields:
                 filters["language"] = toolbar.request_language
             qs = qs.filter(**filters)
             prefetch_cache = {versionable.grouper_field.remote_field.name: qs}
-            related_field._prefetched_objects_cache = prefetch_cache
+
+            # TODO refine it after understand prefetch in many2many field.
```
Aiky30, line 45: I'll take a look at this too.

Aiky30:
@joshyu  There are Flake8 and Isort failures.

djangocms_versioning/plugin_rendering.py
joshyu: @Aiky30  @jonathan-s ,
Can you give me some advice on it. how to handle this?

djangocms_versioning/plugin_rendering.py
joshyu: Thanks mate

djangocms_versioning/plugin_rendering.py
joshyu: > @joshyu There are Flake8 and Isort failures.

I will fix them.

tests/test_monkeypatch.py
joshyu: I will move to a separate test_** file

djangocms_versioning/plugin_rendering.py
jonathan-s: Instead of using `all()` it's probably better to use `get_queryset()` that should preserve any filters that had been added prior to this.

djangocms_versioning/plugin_rendering.py
```diff
@@ -24,14 +27,30 @@ def prefetch_versioned_related_objects(instance, toolbar):
...
+            # because if `related_field` is ManyRelatedManager, it is temporary,
+            # we can't store `_prefetched_objectes_cache` to it, so I decide to store the
+            # prefetched value to model instance.
```
Aiky30, line 48: I'm thinking to prove this we will need to test how many queries are being performed without this and then with it to see if it actually does anything. I'm tempted to change candidate fields to ignore Many to Many relations.

We will need to create a performance test to prove this change to be sure that we are actually adding performance here and not  making it slower by performing a query that isn't actually used. 

self.assertNumQueries allows you to count how many queries are being performed. Cretae a page with many plugins, with and without M2M poll plugin and see if this value is actually changed by this change.

tests/test_plugin_rendering.py
```diff
@@ -0,0 +1,61 @@
+from cms.api import add_plugin
```
Aiky30, line 1: You've done a good job with this, I thought that your issue was a field issue in GDPP, you;re right that this is the issue and the tests that you have prove this well!

djangocms_versioning/plugin_rendering.py
joshyu: @Aiky30 , @chaosk , @jonathan-s ,
I ever browser django cms source code but I haven't seen any place which possibly adopt this cache to improve the query performance. so Actually I don't understand why @chaosk 
 added this field-level prefetched_objects_cache.

Aiky30: -@joshyu This is a Django feature called prefetch_related, that function uses prefetched_objects_cache for query speed ups. You should be searching Django not Django CMS: https://github.com/django/django/blob/93ed71e05802a47774b52503cdc3442686d686c1/django/db/models/base.py#L603

djangocms_versioning/plugin_rendering.py
joshyu: @Aiky30 , @jonathan-s ,
Yes, I tried Jonathan's solution and it works.

djangocms_versioning/plugin_rendering.py
joshyu: @Aiky30 , @jonathan-s ,
I need some more time to finish this test.
I found some resources: 
https://medium.com/@lucasmagnum/djangotip-select-prefetch-related-e76b683aa457
https://github.com/django/django/blob/a34cb5a6d408203f4fbdb364fc9768c026eda224/django/db/models/fields/related_descriptors.py#L907

tests/test_plugin_rendering.py
joshyu: Thanks

djangocms_versioning/plugin_rendering.py
joshyu: @Aiky30  @jonathan-s ,
I update the test and now It can test the cache with 1000 relation access, but all hit the cache.

djangocms_versioning/plugin_rendering.py
```diff
@@ -24,14 +27,29 @@ def prefetch_versioned_related_objects(instance, toolbar):
...
+                filters = {versionable.grouper_field_name: related_field}
             # TODO Figure out grouping values-awareness
             # for extra fields other than hardcoded 'language'
             if "language" in versionable.extra_grouping_fields:
                 filters["language"] = toolbar.request_language
             qs = qs.filter(**filters)
-            prefetch_cache = {versionable.grouper_field.remote_field.name: qs}
-            related_field._prefetched_objects_cache = prefetch_cache
+
+            # TODO refine it after understand prefetch in many2many field.
```
Aiky30, line 44: Remove the TODO, You can add a comment describing what this is doing though

tests/test_plugin_rendering.py
```diff
@@ -0,0 +1,70 @@
...
+        plugin = add_plugin(
+            placeholder, "PollManyPlugin", version.content.language
+        )
+
+        plugin.polls.add(poll)
+        plugin._prefetched_objects_cache = {}
+
+        with self.assertNumQueries(1):
+            contentRenderer.render_plugin(plugin, context)
+            for i in range(1000):
```
Aiky30, line 39: I see what you are trying to do but 1000 is too high, bring this down to 10 or a 100 at most.

tests/test_plugin_rendering.py
```diff
@@ -0,0 +1,70 @@
...
+                self.assertEqual(1, len(plugin.polls.all()))
+
+        with self.assertNumQueries(0):
+            self.assertEqual(1, len(plugin.polls.all()))
```
Aiky30, line 43: This test is now redundant, its covered in the loop above.

tests/test_plugin_rendering.py
```diff
@@ -0,0 +1,70 @@
...
+
+        self.assertIsNotNone(plugin._prefetched_objects_cache)
+        self.assertIsNotNone(plugin._prefetched_objects_cache['polls'])
+        self.assertEqual(len(plugin._prefetched_objects_cache['polls']), 1)
+
+    def test_prefetch_versioned_m2o_objects(self):
+        '''
+        #test model with one foreign key to grouper model
```
Aiky30, line 51: Remove the has in the comment

PR #220, opened 2020-03-31
Enable copy plugins - remove translations is commented out

jonathan-s:
The change_language_menu is taken from the most recent version
in django-cms. In essence we don't need to monkeypatch this,
right now we still need this monkeypatch because the work to
overwrite the delete view in pagecontent admin is not yet done.

Depends on:
- https://github.com/divio/django-cms/pull/6828
- https://github.com/divio/django-cms/pull/6829
- https://github.com/divio/django-cms/pull/6827
- https://github.com/divio/djangocms-versioning/pull/218


Aiky30:
The toolbar is no longer monkey patched to use the toolbar system correctly.

jonathan-s:
@Aiky30 Glad you're slowly working around all these PRs. I do hope they came to use in the end.

Aiky30:
@jonathan-s I've been a little reluctant to change much in versioning right now due to the fact the it's being used in some big sites now. I'm hoping to pick up the work that you did for bringing the comparison improvements to life again.

PR #222, opened 2020-04-22
Fix: Edit button doesn't redirect to the correct editable objects language 

Aiky30:
The issue:
The edit button does not redirect to the correct edit end point in the cms.

The solution:
Provide the language to the edit endpoint if it is available for the object the user wishes to edit!

Steps to replicate:
1. Create a new page in one language (English)
2. Add a translation for the page in another language (Italian)
3. Select the Italian page using the Language selector in the toolbar
4. Click the "Edit" button
5. You will be redirected to edit the correct PageContent but any plugins added to the page will be added in the wrong language!

To be able to replicate this issue you will first need the following fix that allows you to navigate to the preview endpoint of a page: https://github.com/divio/djangocms-versioning/pull/222


SydorenkoSergiy:
Checked the behavior. 
Works properly!

adam-murray:
Also checked functionality and it fixes the issue.

Aiky30:
Merging, pending test has actually completed and passed.

PR #223, opened 2020-04-23
Rework the Page toolbar to use PageContent preview links, the change allows more control in the future

Aiky30:
Fixes an issue where trying to edit draft content using the language selector in the toolbar tries to redirect the user to the live URL which is not yet published, we should always use the preview url as this endpoint can cater for all eventualities.

Reworked the page toolbar to override it rather than hacking it via a monkey patch, this also gives versioning much more control over the toolbar for future upcoming changes.

The issue: When using the toolbar "Language" menu to navigate to a page that is a draft i.e. never been published, the URL is the live URL of the page, whereas it should use the placeholder preview endpoint that will always render! 

Steps to replicate:
  1. Create a new draft page in one language
  2. View the page with the toolbar enabled
  3. Select the toolbar drop down selector labelled "Language"
  4. Click on a Language in the selector


Divio-Test:
Checked. The issue is fixed

SydorenkoSergiy:
Checked. The issue is fixed

djangocms_versioning/cms_toolbars.py
```diff
@@ -131,6 +144,87 @@ def post_template_populate(self):
         self._add_versioning_menu()
 
 
+class VersioningPageToolbar(PageToolbar):
+    """
+    Overriding the original why???
+
+    The default language menu contains the
```
adam-murray, line 151: Incomplete comment.

djangocms_versioning/cms_toolbars.py
```diff
@@ -131,6 +144,87 @@ def post_template_populate(self):
...
+                return None
+
+            languages = get_language_dict(self.current_site.pk)
+
+            remove = [
+                (code, languages.get(code, code))
+                for code in self.page.get_languages()
+                if code in languages
+            ]
+            add = [l for l in languages.items() if l not in remove]
```
adam-murray, line 210: At first glance it is hard  to tell whether these are l's or 1's, could use lang or something instead?

djangocms_versioning/cms_toolbars.py
```diff
@@ -131,6 +144,87 @@ def post_template_populate(self):
...
+            # remove_item uses `items` attribute so we have to copy object
+            for _item in copy(language_menu.items):
+                language_menu.remove_item(item=_item)
+
+            for code, name in get_language_tuple(self.current_site.pk):
+                # Get the content, it could be draft too hence using _original_manager!
+                page_content = self.get_page_content(language=code)
+                if page_content:
+                    url = get_object_preview_url(page_content, code)
+                    language_menu.add_link_item(name, url=url, active=self.current_lang == code)
```
jonathan-s, line 188: This method replaces the old urls with new urls that are guaranteed to be able to fetch drafts. I don't think this would invalidate the following PR: https://github.com/divio/django-cms/pull/6826

I would  say that the testcase I added in that PR should pass because that is a behaviour we want even without versioning. So just keeping that in mind. 

The same goes for this PR: https://github.com/divio/django-cms/pull/6828 the links should be correct without versioning as well.

djangocms_versioning/cms_toolbars.py
```diff
@@ -131,6 +144,87 @@ def post_template_populate(self):
...
+
+    def change_language_menu(self):
```
jonathan-s, line 190: I'm noting here that this is just moving old code to this new place. When this all over and done I would feel inclined to say that this method should be inherited from PageToolbar.

djangocms_versioning/cms_toolbars.py
Aiky30: I think this PR would mean the following is no longer required because there is no such thing as draft in standalone: https://github.com/divio/django-cms/pull/6826/files


I think that this PR is still valid because the url generated is always using the currnet page and not the language inside the loop: https://github.com/divio/django-cms/pull/6828/files

djangocms_versioning/cms_toolbars.py
Aiky30: I agree, the main issue in the core is that this method is added in the Basic Toolbar, this toolbar is separate to all of the others which makes accessing and overriding it difficult. I'm not sure that the basictoolbar should have this method at all! For now this is the least invasive way.

djangocms_versioning/cms_toolbars.py
jonathan-s: Regarding 6826, I'd argue for the sake of consistency to keep it the same; ie preview-url in core, preview-url in versioning.

djangocms_versioning/cms_toolbars.py
jonathan-s: It was merely an observation, this is completely fine. I made the observation to confirm that nothing else had been changed as well.

djangocms_versioning/cms_toolbars.py
Aiky30: Ok, I agree on that for consistency. I'll comment in that PR so it's not lost.

PR #225, opened 2020-05-11
Release 0.0.27

Aiky30:


Aiky30:
All checks have passed, seems to be an issue with the git hook as it still appears to be pending. Re-ran the workflow in Circle CI to try and kick the hook back into life.

PR #232, opened 2020-12-15
Changes added to display Content version number along with pk in compare view

NarenderRajuB:
Changes added to display Content version number instead verion pk in compare view and also updated local timezone date display in compare
Current
image

Updated
image

### DIFF
--- a/djangocms_versioning/templates/djangocms_versioning/admin/compare.html
+++ b/djangocms_versioning/templates/djangocms_versioning/admin/compare.html
@@ -3,7 +3,20 @@
 <html lang="en">
 <head>
     <meta charset="UTF-8">
-    <title>Compare {{ v1 }} {{ v1.created }}{% if v2 %} to {{ v2 }} {{ v2.created }}{% endif %}</title>
+    <title>{% spaceless %}
+        {% if v1 and v2 %}
+        {% blocktrans with left=v1.description right=v1.description %}
+        Compare {{ left }} to {{ right }}
+        {% endblocktrans %}
+    {% elif v1 %}
+        {% blocktrans with left=v1.description %}
+        Compare {{ left }}
+        {% endblocktrans %}
+    {% elif right %}
+        {% blocktrans with right=v2.description %}
+        Compare {{ right }}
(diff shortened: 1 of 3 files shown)


Aiky30:
@NarenderRajuB I think the change is supposed to be that the version pk is not shown to the user, they are expecting a version number like: 1, 2, 3, 4 ... not the id. The drop down and the UI should only show the version number and not the pk. The url will still need to have the version pk because it needs that for the compare view to work. :-)

NarenderRajuB:
> @NarenderRajuB I think the change is supposed to be that the version pk is not shown to the user, they are expecting a version number like: 1, 2, 3, 4 ... not the id. The drop down and the UI should only show the version number and not the pk. The url will still need to have the version pk because it needs that for the compare view to work. :-)

-- Updated the PR with changes and Attached current and updated compare view screenshots

djangocms_versioning/templates/djangocms_versioning/admin/compare.html
```diff
@@ -13,23 +26,23 @@
             "v2_url": "{{ v2_preview_url }}",
         {% endif %}
         {% if v2 %}
-        "v2_description": "{{ v2 }} {{ v2.created }}",
+            "v2_description": "{{ v2_description }}",
```
Aiky30, line 29: It's disappointing to see that when this has been changed that no tests have failed, including the fact that we are using the version number rather id now. 

We need a test that checks that the correct values are rendered in the compare view, this will help us refactor the compare view in future too because we will be checking that the values are rendered correctly. Any change to this logic in future should fail a test.

Aiky30 (changes requested):
All looks really good mate. The only thing that I was hoping is that a test already exists and would have failed. 

Can you please create a test that checks the description is rendered as we expect for each version?

djangocms_versioning/templates/djangocms_versioning/admin/compare.html
NarenderRajuB: Added test for the changes

Aiky30 (approved):

PR #234, opened 2021-08-04
feat: Generic reusable admin mixin

adam-murray:
Generic extended admin mixin for versioning controls within  content models.


djangocms_versioning/admin.py
```diff
@@ -135,6 +141,176 @@ def has_change_permission(self, request, obj=None):
...
+
+    def get_list_display_configuration(self):
+        """
+        Return the configured list display from cms config versionable object, if it is not configured,
+        return the default django value if list_display is not set.
+        """
+        versionable = versionables.for_content(self.model)
+        return versionable.admin_list_display_fields or ["__str__"]
+
+    def get_preview_method_configuration(self):
```
Aiky30, line 200:
```suggestion
    def _get_preview_url(self, obj):
```

This is getting a url not returning a function / class method.

djangocms_versioning/admin.py
```diff
@@ -135,6 +141,176 @@ def has_change_permission(self, request, obj=None):
...
+        """
+        Return the preview method if available, otherwise return None
+        :return: method or None
+        """
+        versionable = versionables.for_content(self.model)
```
Aiky30, line 205: I don't think that this will work, it needs the actual object instance, I'm not sure if model will be the actual instance or the raw class.  Maybe send int he obj always to be sure, make it a reusable method too.

djangocms_versioning/admin.py
```diff
@@ -135,6 +141,176 @@ def has_change_permission(self, request, obj=None):
...
+
+    def _get_preview_link(self, obj, request, disabled=False):
+        """
+        Return a user friendly button for previewing the content model
+        :param obj: Instance of versioned content model
+        :param request: The request to admin menu
+        :param disabled: Should the link be marked disabled?
+        :return: Preview icon template
+        """
+        preview_url = self.get_preview_method_configuration()
```
Aiky30, line 246:
```suggestion
        preview_url = self._get_preview_url(obj)
```
If accepting the changes above ^^^

djangocms_versioning/helpers.py
```diff
@@ -306,3 +306,21 @@ def get_latest_admin_viewable_page_content(page, language):
     ).order_by(
         "versions__state"
     ).first()
+
+
+def get_list_display_config(model):
```
Aiky30, line 311: Why does this helper exist? I can't see it used anywhere and no comments as to why it could exist.

djangocms_versioning/admin.py
```diff
@@ -135,6 +141,176 @@ def has_change_permission(self, request, obj=None):
...
+        """
+        Get the last modified date of a version
+        :param obj: Versioned content model Instance
+        :return: Modified Date
+        """
+        return self.get_version(obj).modified
+
+    get_modified_date.short_description = _("Modified")
+
+    def get_list_display_configuration(self):
```
Aiky30, line 192: I'm not sure that we need this because whoever is implementing the Mixin can set which fields they want to show on the admin instance rather than a new configuration. It could just be that the default for list_display_fields = []

Then usage is:
list_display_fields = ["my_field"]

This way it keeps the configuration option with the relevant section, you would only use the mixin if you had a custom admin.

tests/test_admin.py
```diff
@@ -2408,3 +2409,78 @@ def test_change_view_action_compare_versions_three_selected(self):
             response = self.client.post(endpoint, data, follow=True)
 
         self.assertContains(response, "Exactly two versions need to be selected.")
+
+
```
Aiky30, line 2413: We're still missing a lot of the tests for the admin actions. They will need to be ported across from navigation and pageadmin.

Aiky30 (changes requested):

djangocms_versioning/admin.py
adam-murray: Totally agree apart from leaving it empty. The default for Django is to return __str__ if list_display isn't populated, which would make sense if they haven't provided their own.

djangocms_versioning/admin.py
```diff
@@ -248,7 +419,8 @@ def state_actions(obj):
         return state_actions
 
     def get_list_display(self, request):
-        return (
+        list_display = self.list_display
```
Aiky30, line 422: This is for the version table, we should leave this alone.

djangocms_versioning/test_utils/polls/cms_config.py
```diff
@@ -22,6 +22,7 @@ class PollsCMSConfig(CMSAppConfig):
             version_list_filter_lookups={"language": get_language_tuple},
             copy_function=default_copy,
             preview_url=PollContent.get_preview_url,
+            # content_admin_mixin=ExtendedVersionAdminMixin,
```
Aiky30, line 25: Old code that can be cleaned out

djangocms_versioning/static/djangocms_versioning/js/create_url.js
```diff
@@ -0,0 +1,71 @@
+(function($) {
```
Aiky30, line 1: We don't have the BE content type widget, I guess it would be nice to supply the widget but I was hoping to change this implementation in the future where a package can supply its widget to a content type search field.

tests/test_admin.py
```diff
@@ -2408,3 +2408,119 @@ def test_change_view_action_compare_versions_three_selected(self):
...
+        # Check response is valid
+        self.assertEqual(200, response.status_code)
+        # Check list_action links are rendered
+        self.assertContains(response, "cms-versioning-action-btn")
+        self.assertContains(response, "cms-versioning-action-preview")
+        self.assertContains(response, "cms-versioning-action-edit")
+        self.assertContains(response, "cms-versioning-action-manage-versions")
+        self.assertContains(response, "js-versioning-action")
+
+    def test_extended_version_change_get_list_display_without_cms_config(self):
```
Aiky30, line 2457: cms_config is no longer used. Redundant test.

tests/test_admin.py
```diff
@@ -2408,3 +2408,119 @@ def test_change_view_action_compare_versions_three_selected(self):
...
+        # Check list_action links are rendered
+        self.assertContains(response, "cms-versioning-action-btn")
+        self.assertContains(response, "cms-versioning-action-preview")
+        self.assertContains(response, "cms-versioning-action-edit")
+        self.assertContains(response, "cms-versioning-action-manage-versions")
+        self.assertContains(response, "js-versioning-action")
+
+    def test_extended_version_change_list_display_renders_without_list_display(self):
+        """
+        Check the list_display options added by ExtendedVersionAdminTestCase are
```
Aiky30, line 2440:
```suggestion
        A default is set for the content object if the class inheriting the mixin: ExtendedVersionAdminMixin has not set any fields to display.
```

tests/test_admin.py
```diff
@@ -2408,3 +2408,119 @@ def test_change_view_action_compare_versions_three_selected(self):
...
+        populated from cms_config are rendered in change_list
```
Aiky30, line 2441:
```suggestion
```

tests/test_admin.py
```diff
@@ -2408,3 +2408,119 @@ def test_change_view_action_compare_versions_three_selected(self):
             response = self.client.post(endpoint, data, follow=True)
 
         self.assertContains(response, "Exactly two versions need to be selected.")
+
+
+class ExtendedVersionAdminTestCase(CMSTestCase):
+
+    def test_extended_version_change_list_display_renders_from_provided_list_display(self):
+        """
+        Check the list_display options added by ExtendedVersionAdminTestCase
```
Aiky30, line 2417: All fields are present for a content object if the class inheriting the mixin: ExtendedVersionAdminMixin has set any fields to display. This will be the list of fields the user has added and the fields & actions set by the mixin.

tests/test_admin.py
```diff
@@ -2408,3 +2408,119 @@ def test_change_view_action_compare_versions_three_selected(self):
...
+        populated from cms_config are rendered in change_list
+        """
+        factories.BlogContentWithVersionFactory()
+
+        with self.login_user_context(self.get_superuser()):
+            response = self.client.get(self.get_admin_url(BlogContent, "changelist"))
+
+        # Check response is valid
+        self.assertEqual(200, response.status_code)
+        # Check list_action links are rendered
```
Aiky30, line 2450: No default value tested for here?

tests/test_admin.py
```diff
@@ -2408,3 +2408,119 @@ def test_change_view_action_compare_versions_three_selected(self):
...
+        self.assertIn("cms-versioning-action-edit", list_actions)
+        self.assertIn("cms-versioning-action-manage-versions", list_actions)
+        self.assertIn("js-versioning-action", list_actions)
+
+
+class ListActionsTestCase(CMSTestCase):
+    def setUp(self):
+        self.modeladmin = admin.site._registry[PollContent]
+
+    def test_edit_link(self):
```
Aiky30, line 2490:
```suggestion
    def test_edit_link(self):
        """
        The edit link should be shown when a version is editable. A published version can show an edit button which causes a new draft to be created.
        """
```

tests/test_admin.py
```diff
@@ -2408,3 +2408,119 @@ def test_change_view_action_compare_versions_three_selected(self):
...
+
+        func = self.modeladmin._list_actions(request)
+        edit_endpoint = reverse("admin:djangocms_versioning_pollcontentversion_edit_redirect", args=(version.pk,),)
+        response = func(menu_content)
+
+        self.assertIn("cms-versioning-action-btn", response)
+        self.assertIn('title="Edit"', response)
+        self.assertIn(edit_endpoint, response)
+
+    def test_edit_link_inactive(self):
```
Aiky30, line 2505:
```suggestion
    def test_edit_link_inactive(self):
        """
        The edit link should not be shown for a user that does not have the edit permission. 
        """
```

tests/test_admin.py
```diff
@@ -2408,3 +2408,119 @@ def test_change_view_action_compare_versions_three_selected(self):
...
+
+        func = self.modeladmin._list_actions(request)
+        edit_endpoint = reverse("admin:djangocms_versioning_blogcontentversion_edit_redirect", args=(version.pk,),)
+        response = func(version.content)
+
+        self.assertIn("inactive", response)
+        self.assertIn('title="Edit"', response)
+        self.assertNotIn(edit_endpoint, response)
+
+    def test_edit_link_not_shown(self):
```
Aiky30, line 2519:
```suggestion
    def test_edit_link_not_shown(self):
        """
        The edit link should not be shown when no user is logged in.
        """
```

Aiky30: I'm not really sure what this test is doing because the user will have to be logged in to access the admin.

Aiky30 (changes requested):

tests/test_admin.py
Aiky30: This should be removed

Aiky30 (approved):

PR #243, opened 2021-11-16
Feature: Copy method and Monkeypatch CMS Pagecontent Admin Extensions

adam-murray:
Copy method added to allow versioning, as well as monkey patch to override versioning query set filters when saving and adding extensions.

### DIFF
--- a/tests/test_monkeypatch.py
+++ b/tests/test_monkeypatch.py
@@ -21,6 +21,7 @@ from djangocms_versioning.test_utils.factories import (
     PageVersionFactory,
     PlaceholderFactory,
     PollVersionFactory,
+    TestTitleExtensionFactory,
     TextPluginFactory,
     PollTitleExtensionFactory,
 )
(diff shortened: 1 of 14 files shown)


tests/test_monkeypatch.py
```diff
@@ -54,6 +61,75 @@ def test_copy_extensions(self):
...
+        """
+        The page content copy method should create a new title extension, if one is attached to it.
+        """
+        poll_extension = PollTitleExtensionFactory(extended_object=self.pagecontent)
+        poll_extension.votes = 5
+
+        new_pagecontent = copy_page_content(self.pagecontent)
+
+        self.assertNotEqual(new_pagecontent.polltitleextension, poll_extension)
+        self.assertEqual(new_pagecontent.polltitleextension.votes, 5)
```
Aiky30, line 74: We should also check the pk's here as the copy method should show that the original PageContent has the Extension with Pk X and the new copied PageContent has an extension with a Pk that's not the same as before and is a new one.

This shows that a new model was created and enforces that they are different i.e. it's equal to what we expect and enforces not equal to the original.
self.pagecontent.polltitleextension.pk is equal to poll_extension.pk
new_pagecontent.polltitleextension.pk not equal poll_extension.pk

tests/test_monkeypatch.py
```diff
@@ -54,6 +61,75 @@ def test_copy_extensions(self):
...
+        """
+        poll_extension = PollTitleExtensionFactory(extended_object=self.pagecontent)
+        poll_extension.votes = 5
+        title_extension = TestTitleExtensionFactory(extended_object=self.pagecontent)
+
+        new_pagecontent = copy_page_content(self.pagecontent)
+
+        self.assertNotEqual(new_pagecontent.polltitleextension, poll_extension)
+        self.assertNotEqual(new_pagecontent.testtitleextension, title_extension)
+        self.assertEqual(new_pagecontent.polltitleextension.votes, 5)
```
Aiky30, line 99: As above check that the pk's are what you expect and are not the same as the original.

Aiky30 (changes requested):
Good effort. literally 2 minor additions to the tests to ensure that what we have copied is what we expected to, and it doesn't match the original i.e. it actually duplicated / copied the extension.

Aiky30 (approved):

PR #249, opened 2022-01-04
Fix: Field Ordering Added to Generic Versioning Mixin

adam-murray:
Field ordering was missing in the generic admin mixing for state, author and modified date. This PR adds those.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,9 +4,6 @@ Changelog
 
 unreleased
 ==========
-
-0.0.32 (2022-01-05)
-===================
 * fix: Added field ordering to the generic versioning mixing
 
 0.0.31 (2021-11-24)
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,10 @@ Changelog
 unreleased
 ==========
 
+0.0.32 (2022-01-05)
+===================
+* fix: Added field ordering to the generic versioning mixing
+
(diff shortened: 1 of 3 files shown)


djangocms_versioning/admin.py
```diff
@@ -143,6 +143,9 @@ def has_change_permission(self, request, obj=None):
 class ExtendedVersionAdminMixin(VersioningAdminMixin):
     """
     Extended VersionAdminMixin for common/generic versioning admin items
+
+    WARNING: This makes use of ordered fields, if custom ordering is on anything that inherits this,
```
Aiky30, line 147: Warning makes this sound like something damaging, it's more of a caveat. 

```suggestion
    CAVEAT: Ordered fields are added, if custom ordering is added to any models that inherits this Mixin
```

Aiky30 (changes requested):
Needs a changelog entry. Tests would probably be overkill.

Aiky30 (approved):

Aiky30 (changes requested):
No changelog entry

CHANGELOG.rst
```diff
@@ -5,6 +5,10 @@ Changelog
 unreleased
 ==========
 
+0.0.32 (2022-01-05)
```
Aiky30, line 8: Your PR is an individual change not a release PR, the changelog entry should be in unreleased

Aiky30 (approved):

PR #256, opened 2022-02-22
feat: Compare view opening in new tab

Bernardvdv:
Compare view should always open in a new tab


adam-murray (approved):
Looks good!

djangocms_versioning/static/djangocms_versioning/js/compare.js
```diff
@@ -0,0 +1,14 @@
+window.addEventListener('load', function() {
+  document.querySelector('select[name=action]').addEventListener('change', function() {
+    var selectVal = this.options[this.selectedIndex].text;
+    if (selectVal == "Compare versions"){
```
Aiky30, line 4: Wondering if this will work when the user has changed the language where "Change versions will be different.

CHANGELOG.rst
```diff
@@ -8,6 +8,7 @@ Unreleased
 * Django 3.0, 3.1 and 3.2 support added
 * Python 3.5 and 3.6 support removed
 * Django 1.11 support removed
+* Open compare view in new tab
```
Aiky30, line 11: This will break when the branch is brought up to date where the changes above are now released.

Aiky30 (changes requested):

djangocms_versioning/static/djangocms_versioning/js/compare.js
Bernardvdv: @Aiky30 Good point. Tested this and the action dropdowns to not translate but the Go button will be translated. Looking at another way of checking this now

Aiky30: @Bernardvdv The Compare Versions won't be translated yet because no translations have been submitted for versioning strings. One day soon it will start to receive translations.

adam-murray (approved):

marktsavage (approved):

djangocms_versioning/static/djangocms_versioning/js/compare.js
```diff
@@ -0,0 +1,18 @@
+window.addEventListener('load', function() {
+  document.querySelector('select[name=action]').addEventListener('change', function() {
+    let selectVal = this.options[this.selectedIndex].getAttribute("value");
+    const actionsForm = document.getElementById('changelist-form')
+    if (selectVal == "compare_versions") {
+        // Setting the target to blank to ensure the compare view is opened in a new tab
+        actionsForm.setAttribute('target', '_blank');
+        const submitButton = actionsForm.querySelector('[type="submit"]')
+        // Reloading the page to renew the csrf token which prevented multiple version compare tabs from opening
+        submitButton.addEventListener("click", function (e) {
```
Aiky30, line 10: The one issue with this is that every time a user "changes" the drop down value a new click event is bound. If you add a console.log or breakpoint in here you'll see that when you click it will reload the page X amount of times you changed the drop down. 

I've never tried this before but once you replicate the above issue can you see if the "once" option prevents the multiple bindings? https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#parameters

If not an ugly way that doen;t chnage your script much is to have a variable that tracks the state i.e. isBound = true / false.

Aiky30 (changes requested):

djangocms_versioning/static/djangocms_versioning/js/compare.js
```diff
@@ -0,0 +1,28 @@
...
+    if (selectVal == "compare_versions") {
+        if (handlersBound === false) {
+            // Setting the target to blank to ensure the compare view is opened in a new tab
+            actionsForm.setAttribute('target', '_blank');
+            bindHandlers(submitButton);
+            handlersBound = true
+        }
+    }else {
+        // If the user deselect the compare version action, the "target=_blank" is removed
+        actionsForm.removeAttribute('target');
```
Aiky30, line 25: This will cause an issue where the user selects another action which removes the target attribute, then when they clock the button the target has been undone. The set attribute in this case has to be added everytime the compare versions is selected but not inside the is bound check. 

Apart form that everything else is good.

CHANGELOG.rst
```diff
@@ -4,6 +4,7 @@ Changelog
 
 Unreleased
 ==========
+* Open compare view in new tab
```
Aiky30, line 7:
```suggestion
* feat: Open compare view in new tab
```

Aiky30 (changes requested):

Aiky30 (approved):

PR #257, opened 2022-02-23
Release 1.0.0

Aiky30:
* Python 3.8, 3.9 support added
* Django 3.0, 3.1 and 3.2 support added
* Python 3.5 and 3.6 support removed
* Django 1.11 support removed

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,9 @@ Changelog
 
 Unreleased
 ==========
+
+1.0.0 (2022-02-23)
+==================
 * Python 3.8, 3.9 support added
 * Django 3.0, 3.1 and 3.2 support added
 * Python 3.5 and 3.6 support removed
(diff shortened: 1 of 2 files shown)


marksweb (approved):
Have you got any thoughts on when to start getting these packages on pypi if we've reached 1.x?

Aiky30:
@marksweb I wonder if we introduce a hard block in the first django-cms 4 migration that requires a developer to add a setting for I know what I am doing and I know the risks. Without the setting set you get an error thrown and the database is not migrated. I guess anyone faking the migrations can still get past that, but by faking they haven't actually changed the database structure. One to discuss as a wider group I guess.

marksweb:
@Aiky30 yeah it's a tech meeting agenda item for sure, but I'm just trying to get thoughts started so we have an idea which might become a plan.

I'd like (and I think Mike has mentioned it also) an idea of how we're going to reach official v4. So some kind of road map that starts with releases like this seems a way there. 

An "I know I'm using django-cms 4" setting is probably a fair choice. Could also use a hook on the [pre-migrate](https://docs.djangoproject.com/en/4.0/ref/signals/#pre-migrate) signal to prevent migration. Raise a custom exception if they don't acknowledge it and combine it with a cms version check.

PR #263, opened 2022-03-16
Added js function to hide django messages to prevent overlap

Bernardvdv:


adam-murray (approved):
This approach makes sense to me.

djangocms_versioning/static/djangocms_versioning/js/messages.js
```diff
@@ -0,0 +1,7 @@
+// Hide all messages after timeout to prevent overlap of breadcrumbs
+var djangoMessages = document.getElementsByClassName("messagelist");
+setTimeout(function(){
+    for (var i = 0; i < djangoMessages.length; i ++) {
+        djangoMessages[i].setAttribute('style', 'display:none !important');
+    }
+}, 3000);
```
marktsavage, line 7: Would it be an improvement to put in a fade out, as opposed to setting display none?

Bernardvdv: Good point. Replaced the display none with a fade out effect

Aiky30: This will need to be initiated on document load to ensure that the dom has loaded the messagelist.

Aiky30: This will need to be initiated on document load to ensure that the dom has loaded the messagelist.

djangocms_versioning/admin.py
```diff
@@ -357,7 +357,8 @@ class VersionAdmin(admin.ModelAdmin):
     """
 
     class Media:
-        js = ("admin/js/jquery.init.js", "djangocms_versioning/js/actions.js", "djangocms_versioning/js/compare.js")
+        js = ("admin/js/jquery.init.js", "djangocms_versioning/js/actions.js", "djangocms_versioning/js/compare.js",
+              "djangocms_versioning/js/messages.js")
```
Aiky30, line 361: As much as I agree that this should be in a separate file it does become a headache with packages that have defined their own js files and have duplicated this list and added the packages own scripts too. I would advise moving this to actions and just clearly section this bit with a comment section.

Aiky30, line 361: As much as I agree that this should be in a separate file it does become a headache with packages that have defined their own js files and have duplicated this list and added the packages own scripts too. I would advise moving this to actions and just clearly section this bit with a comment section.

Aiky30 (changes requested):

djangocms_versioning/admin.py
Bernardvdv: Moved to actions.js

Bernardvdv: Duplicate of below

marktsavage (approved):

PR #265, opened 2022-03-31
Feature/add burger menu

marktsavage:
Added support for a burger menu in the actions column on changelist views for any app inheriting from ExtendedVersionAdminMixin or directly importing the relevant media files. Burger menu code taken and modified from djangocms_pageadmin.


djangocms_versioning/static/djangocms_versioning/css/actions.css
```diff
@@ -28,3 +32,77 @@ a.btn.cms-versioning-action-btn.inactive {
...
+extending the pagetree classes provided by CMS
+---------------------------------------*/
+
+/* add shadow on burger menu trigger */
+a.btn.cms-versioning-action-btn:hover, a.btn.cms-versioning-action-btn.open {
+    box-shadow: inset 0 3px 5px rgba(0,0,0,.125);
+}
+
+/* style for each option row */
+ul.cms-pagetree-dropdown-menu-inner li {
```
Aiky30, line 66: The classes are for page-tree but we are an admin flavoured changelist.

djangocms_versioning/static/djangocms_versioning/js/actions.js
```diff
@@ -4,9 +4,13 @@
...
-        // it is not possible to put a form inside a form, so
-        // the versioning actions have to create their own form on click
-        $('.js-versioning-action, .cms-versioning-js-publish-btn, .cms-versioning-js-edit-btn')
+        // INFO: it is not possible to put a form inside a form, so
+        // the versioning actions have to create their own form on click.
+        // Note for any apps inheriting the burger menu, this will also capture those events. 
+        $(` .js-versioning-action,
+            .cms-versioning-js-publish-btn, 
+            .cms-versioning-js-edit-btn, 
+            .cms-page-admin-action-burger-options-anchor`)
```
Aiky30, line 13: I'm not sure we'll need this. We can review once we have a solution.

djangocms_versioning/templates/admin/djangocms_versioning/change_list.html
```diff
@@ -1,2 +1,18 @@
 {% extends "admin/change_list.html" %}
```
Aiky30, line 1: I don't think that we need any of the changes in this file. This is for the page tree.

djangocms_versioning/admin.py
```diff
@@ -355,6 +355,7 @@ def queryset(self, request, queryset):
 class VersionAdmin(admin.ModelAdmin):
     """Admin class used for version models.
     """
+    change_list_template = "admin/djangocms_versioning/change_list.html"
```
Aiky30, line 358: I don't think that we need this. The changes should be added to the [VersionAdminMixin](https://github.com/django-cms/djangocms-versioning/blob/13bfead550c71b78a64c36500039f010f799bcd9/djangocms_versioning/admin.py#L117) , This Mixin is reused by the other 3rd party packages to add the admin actions to the packages. 

The package djangocms-page-admin implementation for the burger cam be left as it is because it is an older implementation pre the mixin which is probably what's confused you.

Aiky30 (changes requested):
Happy to go through these points when you've had the chance to read through.

tests/test_admin.py
```diff
@@ -2521,6 +2521,22 @@ def test_extended_version_change_list_display_renders_without_list_display(self)
...
+        factories.PollVersionFactory(content=content)
+
+        with self.login_user_context(self.get_superuser()):
+            response = self.client.get(self.get_admin_url(PollContent, "changelist"))
+
+        # Check response is valid
+        self.assertEqual(200, response.status_code)
+
+        # Check burger menu class exists in list_actions
+        self.assertContains(response, "cms-versioning-action-btn")
```
Aiky30, line 2538: This will always exist as part of the previous buttons.

Aiky30 (approved):
Good job mate, loaded this in and tested it. Thanks for persisting to get this over, appreciated. :-)

marktsavage:
No worries mate! Good to get it merged.

PR #267, opened 2022-04-13
Adds a "view published" button to the actions of a published piece of content

marksweb:
The only link to a piece of content is the link to the admin preview. There's no obvious or easy way to access the URL belonging to the published version. So this adds an action button to the frontend URL.

I've used the eye icon for this as it seemed most appropriate.

Screenshot 2022-04-13 at 20 46 52


Aiky30 (changes requested):
@marksweb Can you bring in the 2 changes for CI fixes and the new feature separately please? Helps in the blame feature, is good practice and is also good if anything is ever needed to be rolled back via git revert. 

Re for the CI changes, as a separate PR don't see anything additional required to bring that in as is.  

Re for the published action / button, this will need additional work for a test to ensure that when something is versioned and has a get_absolute_url it is shown and works as expected, likewise it is not present for any other versions (Draft, Archived, Unpublished). We may also need to consider that it should not work if the version doesn't have a live accessible url (alias doesn't).

PR #272, opened 2022-04-14
feat: Adds a link to published content

marksweb:
This adds an action button on the published content to take you to the content users will see via `get_absolute_url`.

How it looks with page content;
Screenshot 2022-04-13 at 20 46 52

How it looks with Alias where `get_absolute_url` returns the preview URL;
Screenshot 2022-04-14 at 17 11 21

### DIFF
--- a/djangocms_versioning/admin.py
+++ b/djangocms_versioning/admin.py
@@ -535,7 +535,10 @@ class VersionAdmin(admin.ModelAdmin):
             # Don't display the link if it isn't published
             return ""
 
-        published_url = obj.content.get_absolute_url()
+        try:
+            published_url = obj.content.get_absolute_url()
+        except AttributeError:
+            return ""
 
         return render_to_string(
             "djangocms_versioning/admin/published_icon.html",
--- a/djangocms_versioning/admin.py
+++ b/djangocms_versioning/admin.py
@@ -181,6 +181,16 @@ class ExtendedVersionAdminMixin(VersioningAdminMixin):
         else:
             return None
 
+    def _get_published_url(self, obj):
+        """
+        Return the published url method if available, otherwise return None
+        :return: method or None
+        """
+        try:
+            return obj.content.get_absolute_url()
+        except AttributeError:
+            return None
+
     def _list_actions(self, request):
         """
         A closure that makes it possible to pass request object to
@@ -201,7 +211,7 @@ class ExtendedVersionAdminMixin(VersioningAdminMixin):
 
     def _get_preview_link(self, obj, request, disabled=False):
         """
-        Return a user friendly button for previewing the content model
+        Return a user-friendly button for previewing the content model
         :param obj: Instance of versioned content model
         :param request: The request to admin menu
         :param disabled: Should the link be marked disabled?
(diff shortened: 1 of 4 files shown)


tox.ini
```diff
@@ -8,10 +8,8 @@ skip_missing_interpreters=True
 
 [testenv]
 deps =
-    -r{toxinidir}/tests/requirements/requirements_base.txt
```
Aiky30, line 11: I added this because I thought that the isort and flake8 deps needed it, obviously not so no issues removing this.

djangocms_versioning/admin.py
```diff
@@ -537,6 +537,22 @@ def _get_unpublish_link(self, obj, request, disabled=False):
...
+            if not obj.state == PUBLISHED:
+                # Don't display the link if it isn't published
+                return ""
+
+            published_url = obj.content.get_absolute_url()
+        except AttributeError:
+            return ""
+
+        return render_to_string(
+            "djangocms_versioning/admin/published_icon.html",
```
Aiky30, line 552: An icon for preview already exists here, we shouldn't need to create another one: https://github.com/django-cms/djangocms-versioning/blob/master/djangocms_versioning/templates/djangocms_versioning/admin/preview.html

djangocms_versioning/admin.py
```diff
@@ -537,6 +537,22 @@ def _get_unpublish_link(self, obj, request, disabled=False):
...
 
+    def _get_published_link(self, obj, request):
+        """Helper function to get the link to the published page"""
+        try:
+            if not obj.state == PUBLISHED:
+                # Don't display the link if it isn't published
+                return ""
+
+            published_url = obj.content.get_absolute_url()
+        except AttributeError:
```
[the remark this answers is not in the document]
marksweb: @Aiky30 To be honest, I can't remember now - it's been a bit long to remember what I was doing here! I'll take this out though. So I'll take this out.

djangocms_versioning/admin.py
marksweb: @Aiky30 If I were to use that template, then the text describing the icon would be "preview" but it's a link to view the published page. I'll update my template and move it for consistency, but I think an explicit template makes sense.

djangocms_versioning/admin.py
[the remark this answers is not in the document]
marksweb: @Aiky30 Ok, check the latest test run failures...

```
ERROR: test_action_published_link_not_valid_obj (tests.test_admin.VersionAdminActionsTestCase)
The published content link is not applicable with this object
----------------------------------------------------------------------
Traceback (most recent call last):
  File "./tests/test_admin.py", line 657, in test_action_published_link_not_valid_obj
    actual_action_control = self.version_admin._get_published_link(version, request)
  File "/home/runner/work/djangocms-versioning/djangocms-versioning/djangocms_versioning/admin.py", line 542, in _get_published_link
    if not obj.state == PUBLISHED:
AttributeError: 'Answer' object has no attribute 'state'
```

https://github.com/django-cms/djangocms-versioning/runs/8260551716?check_suite_focus=true

djangocms_versioning/admin.py
```diff
@@ -537,6 +537,19 @@ def _get_unpublish_link(self, obj, request, disabled=False):
             {"unpublish_url": unpublish_url, "disabled": disabled},
         )
 
+    def _get_published_link(self, obj, request):
+        """Helper function to get the link to the published page"""
+        if not obj.state == PUBLISHED:
+            # Don't display the link if it isn't published
+            return ""
+
+        published_url = obj.content.get_absolute_url()
```
fsbraun, line 546: I'd suggest to change this to `getattr(obj.content, "get_absolute_url", None)` and assume that not all objects have an absolute url. Then `if not published_url: return ""` to make sure the eye is only shown if there is an url.

fsbraun (changes requested):
Hi @marksweb ! I am not sure if all versionable objects need to have a `get_absolute_url` method.

fsbraun (approved):

PR #280, opened 2022-06-06
fix: Added correct relationship description to get_created_by admin_order_field

adam-murray:
## Description
* In admin menus which make use of the ExtendedVersionAdminMixin, the created_by field was not ordering properly, this is due to the entire relationship not being described in the admin_order_field configuration. By ensuring the last field in the target is set to username, rather than simply the User model, we can sort alphabetically. 

## Checklist

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,7 +6,6 @@ Unreleased
 ==========
 * fix: Added correct relationship description to get_created_by admin_order_field
 
-
 1.2.0 (2022-06-09)
 ==================
 * feat: Add View Published button for page edit or preview mode
@@ -14,7 +13,6 @@ Unreleased
 1.1.0 (2022-06-08)
 ==================
 * feat: Added injection point for field modification in the ExtendedAdminMixin
-
 
 1.0.6 (2022-05-31)
 ==================
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,17 @@ Changelog
 Unreleased
 ==========
 * fix: Added correct relationship description to get_created_by admin_order_field
+
+
+1.2.0 (2022-06-09)
+==================
+* feat: Add View Published button for page edit or preview mode
+
+1.1.0 (2022-06-08)
+==================
+* feat: Added injection point for field modification in the ExtendedAdminMixin
+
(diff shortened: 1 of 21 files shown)


CHANGELOG.rst
```diff
@@ -4,7 +4,7 @@ Changelog
 
 Unreleased
 ==========
-
+* fix: Added correct relationship description to get_created_by admin_order_field
 1.0.6 (2022-05-31)
```
Bernardvdv, line 8:
```suggestion

1.0.6 (2022-05-31)
```

Bernardvdv (approved):

tests/test_admin.py
```diff
@@ -2708,6 +2708,71 @@ def test_extended_version_change_list_actions_burger_menu_available(self):
...
+            created_by=user_last,
+        )
+        factories.PollVersionFactory(
+            content=factories.PollContentFactory(language="en"),
+            created_by=user_last_lower,
+        )
+
+        with self.login_user_context(self.get_superuser()):
+            base_url = self.get_admin_url(PollContent, "changelist")
+            base_url += "?o=3"
```
Aiky30, line 2750: May need to add one thing to this test to ensure that the index of the 3rd field matches the author field, this will break if anyone adds a new field and someone fixing the test may not know why it's failing. Maintenance of tests is how tests get "patched" and broken forever. The core cms suffers badly from this where half the suite doesn't actually test what it was intended to test anymore. 

To bullet proof it you could get the list display of the admin model and find the index by the field name, that way any changes to fields won't require a manual change.

Aiky30 (changes requested):

tests/test_admin.py
Aiky30:
```
from django.contrib import admin

# Find the index of the author field to sort by the correct field
modeladmin = admin_site._registry[PollContent]
admin_field_list = modeladmin.get_list_display(request)
author_index = admin_field_list.index("get_author")
```

I'm seeing this as index 1 locally with PollContent so not sure if '3' is correct. 3 would be version state: 
Screenshot 2022-06-08 at 16 04 06

Bernardvdv (approved):

Aiky30 (approved):

PR #287, opened 2022-06-25
feat: Django 4.0, 4.1 / Python 3.10/3.11, mysql support, running tests on sqlite, postgres and mysql

### LINKED ISSUE #286: Migration 009 fails on mysql
I just tried to install a minimal django-cms v4 project inlcuding versioning on a mysql database. 

Migration 009 of djangocms-versioning fails with an error: `django.db.transaction.TransactionManagementError: Executing DDL statements while in a transaction on databases that can't perform a rollback is prohibited.`

The same initial migration succedes on postgres.

fsbraun:
## Description

This PR extends the test suite considerably by introducing tests for Postgres and MySQL.  It is a fix for #286 where a migration that fails to run on MySQL. 

It turns out that some unit tests did neither run on Postgres nor MySQL, since they compared pk of Version objects with pk of the corresponding Content object which only accidentally worked on sqlite. 

## Features of the PR

* Extended tests running on sqlite, Postgres, MySQL
* Added tests for Django 4.0, 4.1, Python 3.10, 3.11
* Fix for #286

## Checklist


Aiky30:
@fsbraun Can you add more details to the error / issue such as MySQL version and the database backend in use, for postgres this tends to be: `django.db.backends.postgresql`, although this is up to the project author:   https://docs.djangoproject.com/en/4.1/ref/databases/#postgresql-connection-settings

I ask because this chnage doesn't single out MySQL and affects all DB's. Ideally we would try and avoid that.

fsbraun:
I've since switched this instance to postgres. But I used the mysql backend: `django.db.backends.mysql`.  Also set was

        DATABASES["default"]["OPTIONS"] = {
            "init_command": "SET sql_mode='STRICT_TRANS_TABLES'"
        }


`mysqld --version` yields `5.7.38` using an `InnoDB` database. (On another server I use `Distrib 10.5.15-MariaDB` but I did not check there.)

If needed I can setup a new empty project with mysql (which is what I was doing) and give you the full stack trace. To me it seems a rollback capability issue with MySql. 

Let me know if I should set up an empty project.

tests/requirements/requirements_base.txt
```diff
@@ -10,7 +10,9 @@ mock
 pillow
 pyflakes>=2.1.1
 python-dateutil
+mysqlclient==2.0.3
+psycopg2
 
 # Unreleased django-cms 4.0 compatible packages
-https://github.com/django-cms/django-cms/tarball/develop-4#egg=django-cms
-https://github.com/divio/djangocms-text-ckeditor/tarball/support/4.0.x#egg=djangocms-text-ckeditor
+git+https://github.com/django-cms/django-cms@develop-4#egg=django-cms
```
Aiky30, line 17: It's fine to change these, they were in the previous format because the Divio Cloud didn't like Pip urls, it's was more convenient to use the same urls across the projects i..e in and out of the Divio Cloud.

djangocms_versioning/test_utils/extended_polls/migrations/0001_initial.py
```diff
@@ -0,0 +1,28 @@
+# Generated by Django 4.1.3 on 2022-11-08 17:46
```
Aiky30, line 1: I don't think that the migrations are actually required for test packages, maybe this is a side affect of the testing issue not using postgres. Doesn't really matter to be fair.

Aiky30 (approved):
Looks OK to me. Nothing blocking.

Some comments below. 

The python version list maybe a little extreme. this is very wasteful on resources (not environmentally friendly). As this is Django 4.1 I guess we need to support what django do, so we could remove 3.7 at least. https://docs.djangoproject.com/en/4.1/releases/4.1/#python-compatibility

tests/requirements/requirements_base.txt
fsbraun: Was not aware of this. Will revert.

PR #291, opened 2022-07-27
feat: Compat with cms page content extension changes

marksweb:
This brings compatibility for the current cms v4 setup and the changes I've introduced with;

https://github.com/django-cms/django-cms/pull/7369

And I've included some basic pre-commit hooks for linters to help avoid issues at the point of github actions.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,26 @@
+ci:
+  autofix_prs: false
+
+repos:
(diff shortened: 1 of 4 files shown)


marksweb:
That check failure can't be avoided. The extra code is going to make the coverage drop, but I guess it's a temporary change.

fsbraun (approved):

PR #305, opened 2023-01-10
feat: add localization and transifex support

fsbraun:
## Description

This PR adds localization and transifex support. 

* Transifex config is found in `.tx/config`
* German and English localization files are added (more to come from transifex)
* Quotes in three templates adjusted so that `./manage.py makemessages --all` creates the correct strings

## Related resources

* #...
* #...

## Checklist


djangocms_versioning/templates/djangocms_versioning/admin/archive_confirmation.html
```diff
@@ -19,7 +19,7 @@ <h4>{% blocktrans %} Version number: {{ version_number }}{% endblocktrans %}</h4
     {% csrf_token %}
     <input  class="button confirm-link js-versioning-keep-sideframe"
             type="submit"
-            value="{% trans 'Yes, I\'m sure' %}">
+            value="{% trans "Yes, I\'m sure" %}">
```
marksweb, line 22: We might be better off using `translate` as the tagname.

`trans` still exists as an alias in django main, but I think it'll get removed at some point.

marksweb (approved):

djangocms_versioning/templates/djangocms_versioning/admin/archive_confirmation.html
fsbraun: True. I've replaced all `{% trans %}` tags with `{% translate %}`

PR #317, opened 2023-02-03
fix: Minor usability improvements

fsbraun:
### Ajax error message in page tree

Since django CMS 4.1.0rc1 the page tree shows error messages returned by ajax post requests such as toggling navigation status of a page content object. 

No additional message needs not to be sent to the front end. It only results in the error message shown again after a reload.

### Edit button renamed to "New Draft" if the action creates a new draft (as opposed to go into editing mode of an existing draft)

To avoid accidental creation of many draft objects by clicking the Edit button in the toolbar this PR renames the Edit button to "New Draft" if by clicking it a new draft will be created.
image

### Menu item "Discard changes" in versioning menu

This replaces v3's "Revert to published" function by calling the discard view (without the need of going through "Manage versions")
image

### Replace deprecated icons
* `cms-icon-copy`-> `cms-icon-manage-version`
* `cms-icon-layers` -> `cms-icon-compare`
* Use newly provided `cms.icons.css` resource

## Related resources

* https://github.com/fsbraun/django-cms/blob/1b79a030dbc9b09543d0e7cbe77143397b8ab70a/cms/static/cms/js/modules/cms.pagetree.js#L905-L910
* https://github.com/django-cms/django-cms/pull/7494

## Checklist

### DIFF
--- a/djangocms_versioning/cms_config.py
+++ b/djangocms_versioning/cms_config.py
@@ -1,7 +1,6 @@
 import collections
 
 from django.conf import settings
-from django.contrib import messages
 from django.contrib.admin.utils import flatten_fieldsets
 from django.core.exceptions import (
     ImproperlyConfigured,
--- a/djangocms_versioning/cms_config.py
+++ b/djangocms_versioning/cms_config.py
@@ -358,7 +358,7 @@ class VersioningCMSPageAdminMixin(indicators.IndicatorStatusMixin, VersioningAdm
         try:
             version.check_modify(request.user)
         except ConditionFailed as e:
-            self.message_user(request, force_str(e), messages.ERROR)
+            # Send error message
             return HttpResponseForbidden(force_str(e))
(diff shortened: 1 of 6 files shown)


djangocms_versioning/cms_config.py
```diff
@@ -358,7 +357,7 @@ def change_innavigation(self, request, object_id):
         try:
             version.check_modify(request.user)
         except ConditionFailed as e:
-            self.message_user(request, force_str(e), messages.ERROR)
+            # Send error message
```
marksweb, line 360: What's the change here - don't message the user because they're getting a 403 response anyway?

fsbraun: Indeed. The 403 response triggers a clear message. Sending another one leads to the 403 message being shown - letter for letter the same as the previously sent - once more out of context with the next request response cycle.

marksweb (approved):

PR #320, opened 2023-02-16
fix: deletion of version objects blocked by source fields

fsbraun:
## Description

`source` fields refer to Version objects a version was originally derived (copied) from. They are declared with `on_delete = models.PROTECT`. This blocks deletion of versioned objects (such as pages) **which have sources**.

If `DJANGOCMS_VERSIONING_ALLOW_DELETING_VERSIONS` is set to `True`,  `on_delete = models.SET_NULL` removes the block and a user with permissions to delete Version objects can delete pages. 

Remark: This does not affect the `models.PROTECT` setting for the `Version.content`.

## Related resources

* This change in model is reflected in the existing migrations but does **not affect database constraints** ([see Django docs](https://docs.djangoproject.com/en/dev/ref/models/fields/#django.db.models.ForeignKey.on_delete)). Toggling the settings will therefore not create migrations.
* A **test for missing migrations** has been added. It needs to run as the first test for otherwise versioning's proxy models might create an incorrect migration.
* This test revealed that two verbose names of fields were not referred to in the migrations. This is fixed.

## Checklist


djangocms_versioning/models.py
```diff
@@ -21,6 +21,13 @@
     emit_content_change = None
 
 
+def allow_deleting_versions(collector, field, sub_objs, using):
+    if getattr(settings, "DJANGOCMS_VERSIONING_ALLOW_DELETING_VERSIONS", False):
```
Aiky30, line 25: This setting should be defined in conf.py: https://github.com/django-cms/djangocms-versioning/blob/master/djangocms_versioning/conf.py

This is probably a function that should live in helpers.py, it can then be reused again in the package.

docs/settings.rst
```diff
@@ -0,0 +1,23 @@
+Settings for djangocms Versioning
+=================================
+
+
+.. py:attribute:: DJANGOCMS_VERSIONING_ALLOW_DELETING_VERSIONS
+
+    Defaults to ``False``
+
+    This setting controlls if the ``source`` field of a ``Version`` object is
```
Aiky30, line 9:
```suggestion
    This setting controls if the ``source`` field of a ``Version`` object is
```
Minor typo

docs/settings.rst
```diff
@@ -0,0 +1,23 @@
...
+    protected. It is by default which implies that Django will not allow a user
```
Aiky30, line 10:
```suggestion
    protected. It is protected by default which implies that Django will not allow a user
```

Aiky30 (commented):
Looks good, minor comments added.

Aiky30 (approved):

PR #325, opened 2023-04-07
fix "View Published" and version preview urls for non main language 

evgeny-dmi3ev:
## Description

When I click 'View published' on a non "en-gb" version of a page I'm taken to the "en-gb" version. 
Also when I click to content link in version list for a non "en-gb" version  I'm taken to the "en-gb" version. 

## Checklist

[](https://www.django-cms.org/slack) to find a “pr review buddy” who is going to review my pull request.

### DIFF
--- a/djangocms_versioning/admin.py
+++ b/djangocms_versioning/admin.py
@@ -17,7 +17,7 @@ from django.template.response import TemplateResponse
 from django.urls import Resolver404, re_path, resolve, reverse
 from django.utils.encoding import force_str
 from django.utils.html import format_html, format_html_join
-from django.utils.translation import gettext_lazy as _
+from django.utils.translation import gettext_lazy as _, override, get_language
 
 from cms.models import PageContent
 from cms.utils import get_language_from_request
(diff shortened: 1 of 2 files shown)


djangocms_versioning/admin.py
```diff
@@ -563,13 +563,15 @@ def nr(self, obj):
...
-        url = get_preview_url(content)
-
-        return format_html(
-            '<a target="_top" class="js-versioning-close-sideframe" href="{url}">{label}</a>',
-            url=url,
-            label=content,
-        )
+        language = content.language if hasattr(content, 'language') else get_language()
+        with override(language):
+            url = get_preview_url(content)
```
fsbraun, line 568:
```suggestion
            url = get_preview_url(content, language=language)
```
When passing the language argument to `get_preview_url` you can skip the `with` block.

fsbraun (changes requested):
Hi @evgeny-dmi3ev ! Thanks for this PR! 

I think the changes to `cms_toolbar.py` are not necessary any more since https://github.com/django-cms/django-cms/pull/7519 has been merged. 😀

The changes in `admin.py` can be simplified a bit. Also, I would **appreciate a test** for them. See tests added for https://github.com/django-cms/django-cms/pull/7519 as an example. This way we make sure such a bug does not sneak in some time again!

PR #326, opened 2023-04-26
feat: Django 4.2, Django CMS 4.1.0rc2 compatibility, and version locking

fsbraun:
## Description

This PR adds tests for Django 4.2 and mixins for django CMS' `GrouperModelAdmin` class (new in 4.2rc2). It includes the version locking functionality. Version locking is activated by setting `DJANGOCMS_VERSIONING_LOCK_VERSIONS = True` in the project settings.

Documentation has been updated and tests have been added. 

To stay DRY and avoid code repetition, the `djangocms_versioning.admin` module has been slightly refactored. `ExtendedListDisplayMixin` encapsulates the extension of the change list view of versioned objects using django CMS 4's new config feature. Both versioned grouper admins and versioned content admins inherit it.

Djangocms Versioning now supports both versioned grouper model admins and versioned content model admins. The latter are backwards compatible.

## Related resources

* Fixed a bug where the edit button was not grayed-out if the user did not have edit rights
* adds ruff as linter (see also #329)
* https://github.com/FidelityInternational/djangocms-version-locking

## Checklist

* [x] I have opened this pull request against ``master``
* [x] I have added or modified the tests when changing logic
* [x] I have followed [the conventional commits guidelines](https://www.conventionalcommits.org/) to add meaningful information into the changelog
* [x] I have read the [contribution guidelines ](https://github.com/django-cms/django-cms/blob/develop/CONTRIBUTING.rst) and I have joined #workgroup-pr-review on 
[Slack](https://www.django-cms.org/slack) to find a “pr review buddy” who is going to review my pull request.

PR #329, opened 2023-05-05
ci: Switch flake8 and isort for ruff

marksweb:
This switches our linting from flake8 and isort to ruff.

### DIFF
--- a/djangocms_versioning/helpers.py
+++ b/djangocms_versioning/helpers.py
@@ -2,15 +2,14 @@ import copy
 import warnings
 from contextlib import contextmanager
 
+from cms.toolbar.utils import get_object_edit_url, get_object_preview_url
+from cms.utils.helpers import is_editable_model
+from cms.utils.urlutils import add_url_parameters, admin_reverse
 from django.contrib import admin
 from django.contrib.contenttypes.fields import GenericRelation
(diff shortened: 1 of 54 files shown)


djangocms_versioning/helpers.py
```diff
@@ -83,6 +82,7 @@ def register_versionadmin_proxy(versionable, admin_site=None):
                 versionable.version_model_proxy
             ),
             UserWarning,
+            stacklevel=2
```
marksweb, line 85: @fsbraun I don't know too much about this `stacklevel` attribute, but this is the only change that I can see here that's anywhere near the test failure.

And the test failing runs this function to try to register a versionable model with admin (I think that's the idea here at least). That test code is here;
https://github.com/django-cms/djangocms-versioning/blob/master/tests/test_admin.py#L334

fsbraun: @marksweb I've also added ruff to #326. 

And right, I remember now: When I added `stacklevel=2` (I think CodeQL complained) the `test_admin.py` stated failing. I changed the test which checks if this warning is issued, including comparing the stack level parameter. 

See changed test here: https://github.com/django-cms/djangocms-versioning/blob/eaf5e9d4d1122d8ab6ce09fd20ef827993512667/tests/test_admin.py#L357-L360

PR #343, opened 2023-07-06
Fix 342 hook to copy relations

### LINKED ISSUE #342: PageContentExtension should offer hook copy_relations
If someone extends model `PageContent` by creating a model inheriting from `PageContentExtension`, that extension only creates flat copies. This is because no hook method `copy_relations` is called.

I'm using a model `MySpecialContent` which inherits from `PageContentExtension`. That model has a one-to-one-relation with `PageContent`. `PageContent` is versioned using **djangocms-versioning**. If I create a new version of a CMS page, in the database a new row appears containing a copy of `MySpecialContent`. This is expected behavior.

`MySpecialContent` has some relations which must also be copied whenever I create a new version. CMS3 offered the method `copy_relations` which would perform this action. In CMS3 this method was called implicitly. In CMS4 it however is not called when creating a new copy of a `PageContentExtension`.

I have a patch for this. PR is coming soon.

jrief:
## Description

Details on this can be found in #342.

Running tests on this PR will fail, because the signature of the CMS's [PageContentExtension.copy_relations()](https://github.com/django-cms/django-cms/blob/5caf8d5c2a8e63f523fa9b43fb074adf321a57dd/cms/extensions/models.py#L23) does not fit here.

This has to be changed first in the `devel_4` branch of django-cms itself.

## Related resources

This fixes #342

## Checklist

Unit tests are pending. See above for the reason.

### DIFF
--- a/djangocms_versioning/cms_config.py
+++ b/djangocms_versioning/cms_config.py
@@ -238,7 +238,9 @@ def copy_page_content(original_content):
                 if field.name not in (PageContent._meta.pk.name, "extended_object")
             }
             extension_fields["extended_object"] = new_content
-            field.related_model.objects.create(**extension_fields)
+            new_extension = field.related_model.objects.create(**extension_fields)
+            if callable(getattr(new_extension, 'copy_relations', None)):
+                new_extension.copy_relations(extension)
 
     return new_content
 
--- a/tests/test_cms_config.py
+++ b/tests/test_cms_config.py
@@ -110,7 +110,7 @@ class PageContentVersioningBehaviourTestCase(CMSTestCase):
         new_title = "new slug here"
         data = {
             "title": self.content.title,
-            "slug": new_title
+            "slug": slugify(new_title),
         }
 
         request = req_factory.get("/?language=en")


fsbraun:
@jrief If I look at this it seems that the logic of copying extensions should be invoked by something like
```
    new_extension = old_extension.copy(new_content_object, language=new_content_object.langauge)
```
`copy` will call `copy_relations`.

Already the current versioning code replicates code from the core. 

In a separate PR I'd suggest to remove to deprecate the language parameter since it is not needed. (It's essentially passed on to the `copy_relations` function.)

jrief:
In order to get this functionality working, we have to patch django-cms first and make the language parameter optional.

I'm unsure about the `copy`-method,`….objects.create(**extension_fields)` already copies the object, so what's the intention of a separate `copy` method?

fsbraun:
It’s already [there](https://github.com/django-cms/django-cms/blob/5caf8d5c2a8e63f523fa9b43fb074adf321a57dd/cms/extensions/models.py#L41) , just needs to be called.  Versioning duplicates it incorrectly (.`objects.create`) and will conflict with any extension that overrides it’s copy method.

fsbraun:
Should be fixed by #344. If not, please reopen.

PR #344, opened 2023-07-09
fix: PageContent extension's `copy_relations` method not called

### LINKED ISSUE #342: PageContentExtension should offer hook copy_relations

fsbraun:
## Description

This PR fixes #342 and is based on #343. 

CMS extensions have a [`copy` method](https://github.com/django-cms/django-cms/blob/003104acefed079fc92a0689683fa34cfd01c94f/cms/extensions/models.py#L42-L56). It also takes care of calling `copy_relations`.  

djangocms-versioning needs not recreate this method where it can miss the call of `copy_relations` but instead can use the original method instead.

## Related resources

* #343 
* #342 

## Checklist

### DIFF
--- a/djangocms_versioning/cms_config.py
+++ b/djangocms_versioning/cms_config.py
@@ -1,6 +1,7 @@
 import collections
 
 from cms.app_base import CMSAppConfig, CMSAppExtension
+from cms.extensions.models import BaseExtension
 from cms.models import PageContent, Placeholder
 from cms.utils import get_language_from_request
 from cms.utils.i18n import get_language_list, get_language_tuple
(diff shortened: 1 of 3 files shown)


marksweb (approved):
I've not ran it, but this makes sense 👍

marksweb:
This needs an update as there's conflicts with the main branch.

PR #345, opened 2023-07-24
feat: Improve default copy method to also copy placeholders and plugins

fsbraun:
## Description

Versioning allows specifying a copy method to allow for copying foreign keys, one-to-one fields in a meaningful way for the specific model.

One of the most relevant relations will be placeholders and their plugins. 

To avoid code repetition of largely identical copy methods in any package using djangocms-versioning, I propose to include coyping placeholders in versioned models by default.

This would make custom copy functions obsolete in alias and blog and other content objects to come.

Also, if the verisoned model has a method `copy_relations` it is called on the new object allowing to keep simple copy information within the model.

Tests follow.

## Related resources

* #...
* #...

## Checklist

## Summary by Sourcery

Improves the default copy method to also copy placeholders and plugins, and introduces a mechanism to allow versioned models to define custom copy logic.

Enhancements:
- Extends the default copy method to include placeholders and their associated plugins.
- Adds a mechanism to call a `copy_relations` method on the new object if it exists, allowing models to define custom copy logic.


djangocms_versioning/admin.py
```diff
@@ -1420,7 +1418,8 @@ def changelist_view(self, request, extra_context=None):
             # Check if custom breadcrumb template defined, otherwise
             # fallback on default
             breadcrumb_templates = [
-                f"admin/djangocms_versioning/{breadcrumb_opts.app_label}/{breadcrumb_opts.model_name}/versioning_breadcrumbs.html",
+                f"admin/djangocms_versioning/{breadcrumb_opts.app_label}/"
```
vinitkumar, line 1421: Is this splitting intentional to make the lint Gods happy?

fsbraun: Also, I had to add some spaces that now apparently reduce coverage?!?

vinitkumar: Coverage tools are stupid. Maybe we could increase line-length or explicity increase the line length of this line to bypass this linting error. What do you think?

PR #346, opened 2023-08-07
Add caching to PageContent __bool__

stefanw:
## Description

Add a cache to `PageContent.__bool__`. I've experienced that on a typical CMS edit page there are thousands of calls to this function which all result in a database query.
Obviously this cache could become stale. If you have ideas where that could happen, please share. However, reducing a page load by 2000+ SQL queries sure seems like worth a try.

## Checklist

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,7 @@ Changelog
 Unreleased
 ==========
 * feat: Reversable generic foreign key lookup from version
+* fix: Remove version check when evaluating CMS PageContent objects
 
 2.0.0rc1
 ========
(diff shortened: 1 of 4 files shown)


fsbraun:
@stefanw Some tests are failing. Can you take a look?

stefanw:
I replaced a `__bool__` in the test factory with `is not None` and now the tests fail a bit nicer 😁 . Now it might actually be cache invalidation problems, I'll have a look.

fsbraun:
It is thinkable that you have djangocms-versioning newly installed and there are already page content objects which do not have a version associated. To cover this situation, the database check is necessary.

I think checking for `None` actually is best. `__bool__` now checks that the content object has a version object associated with it. I am not sure if this is used, however. @Aiky30 Do you know more?

PR #353, opened 2023-09-20
ci: Add testing against django main

marksweb:
This adds testing against django's main branch.


fsbraun:
@marksweb You'll need to also swap installation of requirements and django main branch.

fsbraun (approved):

fsbraun:
@marksweb Should we test the Django main branch with django CMS develop-4 ?

fsbraun:
@marksweb Does it make sense to run the py3.12 against develop-4 for the time being?

marksweb:
@fsbraun not if support might take a while.

fsbraun:
I think it's there.

fsbraun:
@marksweb I added the following lines to the three test files that use `assertQuerySetEqual` just after the imports:

```
if not hasattr(CMSTestCase, "assertQuerySetEqual"):
    # Django < 4.2
    CMSTestCase.assertQuerySetEqual = CMSTestCase.assertQuerysetEqual
```

Might be easier to remove later...

PR #356, opened 2023-10-18
Bugfix/use keyword arguments in admin render change form method

vipulnarang95:
## Description

Using keyword arguments in Render change form method for VersioningAdminMixin


fsbraun (approved):
Thanks, @vipulnarang95. Good catch.

PR #357, opened 2023-10-24
fix: Preview link language

fsbraun:
## Description

This PR fixes an issue with the preview links falling back to the primary language.

* No more jumping between languages when clicking preview in the version admin
* Comparing pages not of the primary language works

## Related resources

* #...
* #...

## Checklist

### DIFF
--- a/djangocms_versioning/cms_toolbars.py
+++ b/djangocms_versioning/cms_toolbars.py
@@ -299,7 +299,7 @@ class VersioningPageToolbar(PageToolbar):
         return get_latest_admin_viewable_content(self.page, language=language)
 
     def populate(self):
-        self.page = self.request.current_page or getattr(self.toolbar.obj, "page", None)
+        self.page = self.request.current_page
         self.title = self.get_page_content() if self.page else None
         self.permissions_activated = get_cms_setting("PERMISSION")
 
--- a/djangocms_versioning/cms_toolbars.py
+++ b/djangocms_versioning/cms_toolbars.py
@@ -299,7 +299,7 @@ class VersioningPageToolbar(PageToolbar):
         return get_latest_admin_viewable_content(self.page, language=language)
 
     def populate(self):
-        self.page = self.request.current_page
+        self.page = self.request.current_page or getattr(self.toolbar.obj, "page", None)
         self.title = self.get_page_content() if self.page else None
         self.permissions_activated = get_cms_setting("PERMISSION")
 
--- a/tests/test_admin.py
+++ b/tests/test_admin.py
@@ -388,10 +388,7 @@ class VersionAdminTestCase(CMSTestCase):
         The link returned is the change url for an editable object
         """
         version = factories.PageVersionFactory(content__title="mypage")
-        preview_url = admin_reverse(
-            "cms_placeholder_render_object_preview",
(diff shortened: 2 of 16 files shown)


djangocms_versioning/cms_toolbars.py
```diff
@@ -299,7 +299,7 @@ def get_page_content(self, language=None):
         return get_latest_admin_viewable_content(self.page, language=language)
 
     def populate(self):
-        self.page = self.request.current_page or getattr(self.toolbar.obj, "page", None)
+        self.page = self.request.current_page
```
marksweb, line 302: What was the scenario where the page wasn't on the request and needed to be retrieved from the toolbar?

tests/test_admin.py
```diff
@@ -443,6 +440,14 @@ def setUp(self):
         self.versionable = PollsCMSConfig.versioning[0]
         self.version_admin = admin.site._registry[self.versionable.version_model_proxy]
 
+
+    def test_get_preview_url(self):
+        content = factories.PollContentWithVersionFactory(language="fr")
+        self.assertEqual(
```
marksweb, line 446: I feel like this needs an additional explicit check to verify French is the language of the preview.

djangocms_versioning/cms_toolbars.py
fsbraun: I don't think there is a scenario. I changed it to ensure the identical behaviour as the core (we had several instances where versioning replicated and deviated, leading to problems). This change, however, is not integral to this PR, so I can undo it.
I want to avoid that the page menu appears if a user is at the endpoint of a custom model which has a `page` property.
@marksweb  What's your take: Change or keep?

tests/test_admin.py
fsbraun: @marksweb I reverted this change, since it needs to go into the core (where tests have been adjusted). See https://github.com/django-cms/django-cms/pull/7691/files.

djangocms_versioning/cms_toolbars.py
marksweb: @fsbraun Keep it in in that case 👍

fsbraun: Have reintroduced it!

tox.ini
```diff
@@ -1,8 +1,7 @@
 [tox]
 envlist =
-    flake8
-    isort
-    py{39.310,311}-dj{32,40,41}-sqlite
+    ruff
```
marksweb, line 3: Where is the ruff config? I may be missing something, but can't see it.

fsbraun: @marksweb The ruff config resides in `pyproject.toml`. I just added the ruff dependency to the testing requirements for easier testing. Transition to ruff was in #329

marksweb:
Ok the change here looks good. We just need some further fixes for 3.12 so maybe that should be it's own PR to introduce that and fix the issues so we get a clean run of checks for this?

fsbraun:
@marksweb @vinitkumar The 3.12 issues reside in django-app-helper. From my side we can merge this.

marksweb:
Ok that's fair 👍

marksweb (approved):

marksweb (approved):

Issue #336, opened 2023-05-31
BUG: The option 'cancel changes' within Versioning menu opens up direct and not in popup


svandeneertwegh:
Small bug.

When clicking on 'Cancel changes' here in NL translation it opens direct and not in a popup.
<img width="289" alt="Schermafbeelding 2023-05-31 om 16 05 44" src="https://github.com/django-cms/djangocms-versioning/assets/34129243/b8c6150d-8dbb-4b38-bdfa-3e470dbaa3cd">

fsbraun:
@svandeneertwegh Can I interest you in creating a PR? :-)

PR #360, opened 2023-10-31
#336 bugfix from link_menu_item to modal_menu_item: discard changes link

svandeneertwegh:
## Description

The 'discard changes' link in the versioning menu item (toolbar) is with this PR changed from add_link_item to add_modal_item.

* [v ] I have opened this pull request against ``master``


fsbraun:
@svandeneertwegh I have two observations:
* Upon confirming to discard changes, the modal does **not** close. Instead, the version list of the object is shown in the modal. I think the modal would need to close, and the page should probably redirect to the preview of the published version. 
* It seems to me that the draft version actually is not discarded. I am not sure why. But it appears to stay in the version list shown in the modal.

PR #364, opened 2023-12-29
fix #363: Better UX in versioning listview

### LINKED ISSUE #363: Better UX: Allow max two checked checkboxes when comparing versions
In the versioning list view, one can click on as many checkboxes as he wishes. This then causes a server side validation error, which must be handled. It would be a better user experience if only two checkboxes would be allowed to be selected.

jrief:
## Description

Fixes #363 

Allow max two checked checkboxes when comparing versions.

## Checklist

This pull request is for review and testing. I actually don't know how to add a unit test since it's client functionality only.


fsbraun:
Great idea! I am going to test the UX.

One thought comes to my mind ahead of time: There might be other bulk actions in the future which might require more than 2 versions selected. To make this PR future-proof, would it be possible to check if the option `value="compare_versions"` is the only one?

fsbraun (approved):

PR #393, opened 2024-03-13
fix: Avoid duplication of placeholder checks for locked versions

fsbraun:
## Description

djangocms-versioning adds `default_checks` to placeholders to avoid changing content of published objects. The checks for locked versions were added for every versioned model, but only need adding once.

This PR moves adding the check for locked versions from `cms_config.py` to `apps.py`, where it is only done once.

## Related resources

* #394 
* #...

## Checklist


marksweb (approved):
LGTM 👍

Issue #421, opened 2024-08-23
[bug] Insecure redirect after publishing etc.


fsbraun:
djangocms-versioning implements a set of redirects after a version state change, e.g., redirect to the published version after publishing.

Depending on the site setup, those redirects might fail, e.g. when publishing a child page of an app hooked page. The app hook "swallows" all URLs below its own, and a redirect to the just published page fails with a 404.

djangocms-versioning should first check, if the redirect URLs are reversible and, if not, implement a fallback solution.

Affected IMHO are two instances:
* "Free" redirects like here: https://github.com/django-cms/djangocms-versioning/blob/c9bf9f4d6e0e0fd55a59cab5567e6402daa87dd2/djangocms_versioning/admin.py#L1033
* Published redirects like here (since technically the result of `get_absolute_url` is not guaranteed to be resolvable): https://github.com/django-cms/djangocms-versioning/blob/c9bf9f4d6e0e0fd55a59cab5567e6402daa87dd2/djangocms_versioning/admin.py#L1055-L1057

As a fallback, I suggest using the preview URL if `conf.ON_PUBLISH_REDIRECT in ("preview", "published")`, otherwise the version list url.

To check if the URL is valid, `resolve` can be used:
```
try:
    resolve(url)
except Resolver404:
    url = fallback
```

fsbraun:
@sourcery-ai plan

fsbraun:
@sourcery-ai plan

PR #436, opened 2024-12-06
fix: add resolvability check before redirecting to prevent insecure redirects after publishing

theShinigami:
## Description

After publishing, if there is a requested URL, it first checks if the requested URL is resolvable. If it is not, it defaults to the version list URL.

## Related resources

* #421 

## Checklist

## Summary by Sourcery

Implement a resolvability check for requested URLs in the publish view to prevent insecure redirects after publishing. Update tests to cover scenarios with resolvable and non-resolvable URLs.

Bug Fixes:
- Add a resolvability check for requested URLs before redirecting after publishing to prevent insecure redirects.

Tests:
- Introduce a test to verify the behavior of the publish view when handling resolvable and non-resolvable redirect URLs.

### DIFF
--- a/djangocms_versioning/admin.py
+++ b/djangocms_versioning/admin.py
@@ -1031,7 +1031,6 @@ class VersionAdmin(ChangeListActionsMixin, admin.ModelAdmin, metaclass=MediaDefi
             )
 
         requested_redirect = request.GET.get("next", None)
-
         if conf.ON_PUBLISH_REDIRECT in ("preview", "published"):
             redirect_url=get_preview_url(version.content)
         else:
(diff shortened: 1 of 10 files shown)


djangocms_versioning/admin.py
```diff
@@ -1033,6 +1033,13 @@ def publish_view(self, request, object_id):
...
             redirect_url=get_preview_url(version.content)
+
+            if requested_redirect:
+                try:
+                    resolve(requested_redirect)
+                except Resolver404:
+                    requested_redirect = version_list_url(version.content)
+
         else:
             redirect_url=version_list_url(version.content)
```
fsbraun, line 1044: In the `else:` case the `requested_redirect` is not validated.

fsbraun (changes requested):
Hey @theShinigami ! Thanks for the PR! That's a good step in the right direction.

I have two comments:
* In the case of `conf.ON_PUBLISH_REDIRECT` **not** in `("preview", "published")` the `requested_redirect` is not validated
* Also the published object's absolute url in line 1067 needs to be validated: Some custom object could return an invalid url.

fsbraun (changes requested):
@theShinigami There were two min-comments stuck in the pipeline. Sorry, it took so long to push them out...

fsbraun (approved):

PR #440, opened 2025-01-13
Fix: Unpublished or archived versions not shown in language menu

fsbraun:
## Description

Unpublished or archived languages were not shown in the language menu.

## Related resources

* https://github.com/django-cms/djangocms-versioning/issues/441

* #...

## Checklist

* [x] I have opened this pull request against ``master``
* [x] I have added or modified the tests when changing logic
* [x] I have followed [the conventional commits guidelines](https://www.conventionalcommits.org/) to add meaningful information into the changelog
* [x] I have read the [contribution guidelines ](https://github.com/django-cms/django-cms/blob/develop/CONTRIBUTING.rst) and I have joined #workgroup-pr-review on 
[Slack](https://www.django-cms.org/slack) to find a “pr review buddy” who is going to review my pull request.

## Summary by Sourcery

Bug Fixes:
- Fixed a bug where unpublished or archived languages were not shown in the language menu.

PR #443, opened 2025-01-15
feat: Re-introduce deleting languages of a page

### LINKED ISSUE #441: [BUG] Language chooser in toolbar not showing all options
## Description

In CMS4 with versioning and I18N enabled the following can happen:

Someone creates a page in multiple languages.

The toolbar then shows a language switcher offering all languages that page is available in.

In some situations that language switcher does not show a language, although one or more page content objects exists.

It then happens that neither the language chooser offers the language, nor is it possible  to select that language using the "Add translation" submenu below the "Language" menu.

## Steps to reproduce

Create a page in multiple languages and publish them.

Go to the page in the secondary language.

Open the versioning menu.

Click on the unpublish button. Now all versions of that page content in that secondary language are unpublished.

If you access the page content in the main language, the language chooser now does not allow you to switch to the secondary language, because it is not offered anymore.

The effect then is, that users have no way to access the content they entered.

## Expected behaviour

The language chooser should always offer all languages for which a page content object exists, whether it is published, unpublished, archived, draft, etc.

## Additional information (CMS/Python/Django versions)

recent version of CMS from GitHub, actually commit 4361f0a7553dbd0c3d0aefc72def00fa0531dbb5

recent version of djangocms-versioning from GitHub, actually commit [b191415bc6](https://github.com/django-cms/djangocms-versioning/commit/b191415bc64f05c6e425c5ba53486d1f6f7a12ed)

## Screen shots

Screenshot 2025-01-13 at 15 13 08

## Do you want to help fix this issue?

* [x] Yes, I want to help fix this issue and I will join the channel #pr-reviews on [the Discord Server](https://discord-pr-review-channel.django-cms.org) to confirm with the community that a PR is welcome.
* [ ] No, I only want to report the issue.

### LINKED ISSUE #442: [Feat] Allow to delete added page translation again in Django CMS 4
## Description

After adding a translation to a page ("Language" > "Add translation") in Django CMS 4, there is currently no way how to delete the language version of a page again. Django CMS 3 used to have the option "Language" > "Delete Translation".

## Suggested behavior/feature 

Reintroduce a function "Language" > "Delete Translation" that deletes individual language versions.

Bildschirmfoto 2024-11-21 um 10 21 12

## Do you want to help fix this issue?

* [x]  Yes, I want to help fix this issue by testing a new version before its release.
* [ ] No, I only want to report the issue.

fsbraun:
## Description

If `DJANGOCMS_ALLOW_DELETING_VERSIONS` is set to `True`, the delete translation menu is offered.

fixes #442 

## Related resources

* #442
* #...

## Checklist

## Summary by Sourcery

Re-introduce the ability to delete page translations if `DJANGOCMS_ALLOW_DELETING_VERSIONS` is set to `True`.

Bug Fixes:
- Restore the missing delete translation menu when `DJANGOCMS_ALLOW_DELETING_VERSIONS` is enabled, addressing issue #441.

Enhancements:
- Obtain the correct page content for copying plugins across languages.

## Summary by Sourcery

Restore the "Delete Translation" menu in the language menu if `DJANGOCMS_ALLOW_DELETING_VERSIONS` is set to `True`.

Bug Fixes:
- Restore the missing delete translation menu when `DJANGOCMS_ALLOW_DELETING_VERSIONS` is enabled.

Enhancements:
- Obtain the correct page content for copying plugins across languages.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -19,7 +19,7 @@ jobs:
           python-version: '3.11'
           cache: 'pip'
       - name: Cache dependencies
-        uses: actions/cache@v4.1.2
+        uses: actions/cache@v4.2.0
         with:
           path: ~/.cache/pip
           key: ${{ runner.os }}-pip-${{ hashFiles('docs/requirements.txt') }}
@@ -44,7 +44,7 @@ jobs:
           python-version: '3.11'
           cache: 'pip'
       - name: Cache dependencies
-        uses: actions/cache@v4.1.2
+        uses: actions/cache@v4.2.0
         with:
           path: ~/.cache/pip
           key: ${{ runner.os }}-pip-${{ hashFiles('docs/requirements.txt') }}
--- a/djangocms_versioning/admin.py
+++ b/djangocms_versioning/admin.py
@@ -618,7 +618,9 @@ class VersionAdmin(ChangeListActionsMixin, admin.ModelAdmin, metaclass=MediaDefi
     actions = ["compare_versions", "delete_selected"]
     list_display = (
         "number",
-        "created",
+    ) + (
+        ("created",) if conf.VERBOSE_UI else ()
+    ) + (
         "modified",
         "content",
         "created_by",
@@ -638,6 +640,9 @@ class VersionAdmin(ChangeListActionsMixin, admin.ModelAdmin, metaclass=MediaDefi
 
     class Media:
         js = ["djangocms_versioning/js/versioning.js"]
+
+    def has_module_permission(self, request):
+        return conf.VERBOSE_UI
 
     def get_changelist(self, request, **kwargs):
         return VersionChangeList
@@ -963,10 +968,12 @@ class VersionAdmin(ChangeListActionsMixin, admin.ModelAdmin, metaclass=MediaDefi
         to show versions of.
         """
         language = get_language_from_request(request)
+        versionable = versionables.for_content(self.model._source_model)
         context = dict(
             self.admin_site.each_context(request),
             opts=self.model._meta,
-            form=grouper_form_factory(self.model._source_model, language)(),
+            form=grouper_form_factory(self.model._source_model, language, self.admin_site)(),
+            title=_("Select {} to view its versions").format(versionable.grouper_model._meta.verbose_name),
         )
         return render(request, "djangocms_versioning/admin/grouper_form.html", context)
 
@@ -1038,12 +1045,12 @@ class VersionAdmin(ChangeListActionsMixin, admin.ModelAdmin, metaclass=MediaDefi
 
         if not version.can_be_published():
             self.message_user(request, _("Version cannot be published"), messages.ERROR)
-            return redirect(requested_redirect or redirect_url)
+            return self._internal_redirect(requested_redirect, redirect_url)
         try:
             version.check_publish(request.user)
         except ConditionFailed as e:
             self.message_user(request, force_str(e), messages.ERROR)
-            return redirect(requested_redirect or redirect_url)
+            return self._internal_redirect(requested_redirect, redirect_url)
 
         # Publish the version
         version.publish(request.user)
@@ -1054,9 +1061,25 @@ class VersionAdmin(ChangeListActionsMixin, admin.ModelAdmin, metaclass=MediaDefi
         # Redirect to published?
         if conf.ON_PUBLISH_REDIRECT == "published":
             if hasattr(version.content, "get_absolute_url"):
-                redirect_url = version.content.get_absolute_url() or redirect_url
+                requested_redirect = requested_redirect or version.content.get_absolute_url()
 
-        return redirect(requested_redirect or redirect_url)
+        return self._internal_redirect(requested_redirect, redirect_url)
+
+
+    def _internal_redirect(self, url, fallback):
+        """Helper function to check if the give URL is resolvable
+        If resolvable, return the URL; otherwise, returns the fallback URL.
+        """
+        if not url:
+            return redirect(fallback)
+
+        try:
+            resolve(url)
+        except Resolver404:
+            return redirect(fallback)
+
+        return redirect(url)
+
 
     def unpublish_view(self, request, object_id):
         """Unpublishes the specified version and redirects back to the
--- a/djangocms_versioning/cms_toolbars.py
+++ b/djangocms_versioning/cms_toolbars.py
@@ -36,7 +36,7 @@ from djangocms_versioning.helpers import (
 from djangocms_versioning.models import Version
 
 VERSIONING_MENU_IDENTIFIER = "version"
-CMS_SUPPORTS_DELETING_TRANSLATIONS = version.Version(cms_version) > version.Version("4.1.4") or True
+CMS_SUPPORTS_DELETING_TRANSLATIONS = version.Version(cms_version) > version.Version("4.1.4")
 
 
 class VersioningToolbar(PlaceholderToolbar):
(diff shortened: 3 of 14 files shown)


fsbraun:
@krosefield Can you install this branch to check if it works to your expectations? I'll be adding some tests in the meantime.

filipweidemann:
@fsbraun Hey. I tested this PR together with @krosefield.

First off: the original change request is implemented and works. We now see a "Delete Translations" menu entry and can successfully delete page translations.

However, there is one edge case that might need further refinement. Let me explain:
When deleting a page translation we successfully remove the latest `Version` of that page in the selected language. After doing so we might be left in a weird state though. 

Assume a page with translations in DE & EN, with a versioning history and multiple existing (and previously published) version objects; let's call them versions A, B, C. They were created in that order, all have been published.

Deleting the english page only removes the latest version C. We would expect that the entire versioning history of that page would get deleted as well. We do understand that this is up to debate though, so this is rather our view and can be discarded if you don't agree with us on this one.

But there is a second issue which I can only describe as "inconsistent state" when following the above procedure. Since version C was deleted, we are now left with B as the latest version. Inside the pagecontent admin we also see that this page is marked as "published" with said version. Accessing the German page yields an interesting error: we can't retrieve the english version of the page! Our language switch seems to be unaware of that page from now on, almost as if the page tree is getting disrupted there.

The page DOES exist, but it seems like the `PageUrl` object gets cleared when deleting the page and not created again when falling back to a previous version. I can verify this by clicking the "eye / show page" button inside the admin, entering the page that correctly exists, and then trying to edit the page settings through the cms toolbar. I am greeted with a "DoesNotExist" error page:

swappy-20250115_140625

This seems like a bug to me, the only viable thing you can do to resolve this as a user is to unpublish the english page, delete all versions and recreate the page from the German one.

This was a rather lengthy text, sorry for that, but please ping me if you need any more information. And thank you very much for working on this feature!

fsbraun:
@filipweidemann Thanks for the quick reply. I need to think about deleting all versions. That's a great point. (And maybe the reason it was removed.) 

Concerning your second issue, I believe this is an instance of #441 which was solved in #440. Having a page without published or draft states could shadow other pages/languages.

filipweidemann:
@fsbraun The way I see it is as follows: you have language-scoped pages, each of them with their own version history. That is fine, maybe we have diverging content for some languages, maybe they are maintained by different teams of people, ...

However, when I decide to delete a *whole language*, I would expect that not only the current version is removed but rather the whole language including versions. 

I would expect this simply because in my mind, a page consists of many different languages and their contents, and each language has their own contents (and versions). This almost looks like a graph to me. When deleting a language, I would expect their leaf nodes (versions) to be gone as well. But again, think about it, take your time, this is not necessarily the correct approach for this project. I am in no position to judge this.

Regarding the error: #440 was merged already, I would expect this to already be present in your feature branch, no?

And what about the `PageUrl` error, do you think this is also related to #441? 
I am not sure how this would be fixed by #440 tbh...

Also, why would #441 even trigger here? The admin view / page content view *clearly* tells me that the english page I just deleted is still published. I think this is the other way round, maybe it just *looks* like #441 because the `PageUrl` is gone, therefore the language switcher only outputs one language (DE, the one that still exists / still has a `PageUrl` object).

Or am I missing something here?

filipweidemann:
Oh nvm, you just merged it 3 hours ago. Any chance to update this branch to include #440 so I can test this again? Maybe it does fix it after all..

fsbraun:
@filipweidemann I see your point. After looking into this, I believe this needs to be fixed additionally in the django CMS core:
* The delete confirmation message is a bit garbled
* The code needs to delete **all** content objects for a page-language combination.
I'm afraid, this is not quite such a quick fix as I anticipated.

Next step: Fix translation deleting in the core.

filipweidemann:
@fsbraun thanks for the update! If you need our help testing this again when it is done, just ping us.

fsbraun:
Thank you, @filipweidemann ! https://github.com/django-cms/django-cms/pull/8111 will fix the delete language problem for the upcoming django CMS 4.2. We'll need a backport to django CMS 4.1, too. Once I have that, I will ping you.

fsbraun:
@filipweidemann You could test against this branch and this fix for the core: https://github.com/django-cms/django-cms/pull/8112 

Once I get your feedback I will push a piece of code that will deactivate the Delete Translations menu for versions of django CMS that do not correctly delete **all** versions the page contents. (This will make the menu disappear again until versions are bumped.)

filipweidemann:
@fsbraun we did. Same errors are still present. I'd like to double-down on my hint that a cascaded deletion of the language versions' `PageUrl` is the root cause here.

Let me walk through this with some screenshots. 
This is an excerpt from one of our djangoCMS 4 page content view in German & English. We will be focusing on the "Movies" page.
initial-en
initial-de

As you can see, nothing too fancy here. Let me also show you the initial state of the `PageUrl` instances for that page:
page-urls-before-deletion

Let's try to delete the English language by creating a new version of the German one (we have to do that to enable the edit controls) and then using the new menu controls. We are greeted with this prompt, asking us to confirm: 
deleting-en-prompt

As you can see, not only are the `PageContent`, `Version` & `StateTracking` instances removed, but **also the PageUrl of the English page**! We can confirm this by doing a query again:
page-url-after-deletion

The publishing state of our page inside the admin view now looks like this:
page-state-after-deletion

The page is unpublished, so that's good. But the `PageUrl` is still missing. What if we manually "restore" the last version of that English page? We now get a green "published" bubble but there is still **no** `PageUrl` being created. 

We can access the page though, either by clicking on "view" inside the admin or using the language tabs in the frontend. When we do that, the language chooser will still only display the German language, even when we are obviously on an English page! When we try to edit the pages settings through the frontend editing, we are greeted with this error, confirming that the missing `PageUrl` is probably not the desired state:
error-when-editing-page-settings-in-frontend

All of this can only be replicated when deleting a language whose `PageContent` has **more than one version**. If we just have a single version, this version will be deleted, showing us that the page does not exist, and creating one from the German version yields the correct results.

fsbraun:
Hi @filipweidemann !

Have you installed both #443 and https://github.com/django-cms/django-cms/pull/8112 ?

If a page content object exists in a language,  a page url object must exist.

Deleting a translation implies deleting all page content objects and the page url object. So, if you delete, say, the German translation, the German page url object should go. In the page tree the page should be visible as "not-existing", (an empty circle), **not** unpublished. The fix in the django-cms package should make sure that (besides the page url) **all page content** objects are removed.

I have not been able to reproduce the `DoesNotExist` error. If I follow your steps, my results deviate in that I do **not** see an unpublished German content after deleting the German translation.

filipweidemann:
@fsbraun I manage my deps using `pipenv` so this is my currently active lockfile (omitting other deps):

```
"django-cms": {
    "git": "https://github.com/fsbraun/django-cms",
    "markers": "python_version >= '3.8'",
    "ref": "18c58a16fd29386df736e165b72b062c12f538ac"
},
"djangocms-versioning": {
    "git": "https://github.com/django-cms/djangocms-versioning",
    "ref": "1c11746b655fa189a36b8637adee75a70d239556"
},
```

This was followed by a `docker compose build && docker compose up` to assure that the deps are correctly installed instead of just reusing an already existing container. 

I locked & rebuilt everything again, just to be safe. But I still get this behaviour. Any ideas?

filipweidemann:
Heh this would explain it I guess :D
Let me try again...

filipweidemann:
@fsbraun okay good news: it looks like this works now! I now get a whole bunch of `PageContent` objects being deleted when removing the site, not just one like before. And this is correctly reflected inside the admin.

This is great tbh. I think this is done! Thanks for your effort on this!

vinitkumar (approved):
Looks good to me. Let's get this merged and released 🚢  🚀

PR #444, opened 2025-01-15
fix: Take csrf token from CMS config if possible

### LINKED ISSUE #437: [BUG] StateIndicatorMixin raises CSRF error with CSRF_USE_SESSIONS = True
## Description

I have a standard Grouper/Content model setup (Not dissimilar to djangocms-blog) where the publish/unpublish actions added via the `StateIndicatorMixin` do not work as expected and throw a 403 - CSRF token from POST incorrect.

From my testing this only occurs when we set `CSRF_USE_SESSIONS = True` in `settings.py`. Setting this to `False` gets things working again. Most other settings are reasonable defaults.

EDIT:
After deploying this to our development server it didn't actually fix the issue but now does raise something slightly more useful. With the setting marked as `False` I get

`Uncaught TypeError: document.cookie.match(...) is null` from https://github.com/django-cms/djangocms-versioning/blob/master/djangocms_versioning/static/djangocms_versioning/js/indicators.js#L12

Going back to a localhost instance with the setting as `True`
`Uncaught TypeError: window.top.CMS is undefined` from https://github.com/django-cms/djangocms-versioning/blob/master/djangocms_versioning/static/djangocms_versioning/js/indicators.js#L76

Some other things:
- This doesn't occur for Pages, only custom models which use the Grouper admin mixins.
- By going to "manage versions" instead you can publish or unpublish as expected.

## Steps to reproduce
1. Go to a custom grouper admin. e.g /admin/news/article/
2. Click on state indicator and select publish or unpublish
3. If not in sidebar, nothing happens, otherwise it will show the 403 in a modal.

## Expected behaviour

Page reloads and the object is marked as published/unpublished

## Actual behaviour

403 CSRF error is raised.

## Additional information (CMS/Python/Django versions)

CMS v4.1.1
Django 5.1.3
Djangocms-Versioning 2.1.0

## Snippet
```
@admin.register(Article)
class ArticleAdmin(StateIndicatorMixin, ExtendedGrouperVersionAdminMixin, GrouperModelAdmin):
    content_models = ArticleContent
    extra_grouping_fields = ("language",)
    list_display = ["title", "author", "state_indicator"]
    list_filter = ["author"]
    search_fields = ["author__name"]
    autocomplete_fields = ["author"]
    fieldsets = [...]


@admin.register(ArticleContent)
class ArticleContentAdmin(ExtendedIndicatorVersionAdminMixin, admin.ModelAdmin):
    fieldsets = [(None, {"fields": ["title", "intro", "slug"]}), admin_mixins.META_FIELDSET]

    def get_model_perms(self, request):
        """
        Return empty perms dict to hide the model from admin index.
        """
        return {}
```

## Do you want to help fix this issue?

* [ ] Yes, I want to help fix this issue and I will join the channel #pr-reviews on [the Discord Server](https://discord-pr-review-channel.django-cms.org) to confirm with the community that a PR is welcome.
* [x] No, I only want to report the issue.

fsbraun:
## Description

Bug Fixes:
- Fixed an issue where the CSRF token was not being fetched correctly.

Fixes #437 

## Related resources

* #437 
* #...

## Checklist

### DIFF
--- a/djangocms_versioning/static/djangocms_versioning/js/indicators.js
+++ b/djangocms_versioning/static/djangocms_versioning/js/indicators.js
@@ -20,7 +20,7 @@
                 csrfToken = cookieToken[1];
             }
         }
-        csrfToken = csrfToken || $('input[name="csrfmiddlewaretoken"]').val() || 'no-token';
+        csrfToken = csrfToken || 'no-token';
 
         if (element.attr('target') === '_top') {
             // Post to target="_top" requires to create a form and submit it
--- a/djangocms_versioning/static/djangocms_versioning/js/indicators.js
+++ b/djangocms_versioning/static/djangocms_versioning/js/indicators.js
@@ -14,7 +14,7 @@
         if (window.top.CMS.config.csrf) {
             csrfToken = window.top.CMS.config.csrf;
         } else {
-            const cookieToken = document.cookie.match(/csrftoken=([^;]*);?/)
+            const cookieToken = document.cookie.match(/csrftoken=([^;]*);?/);
 
             if (cookieToken && cookieToken.length > 1) {
                 csrfToken = cookieToken[1];
--- a/djangocms_versioning/static/djangocms_versioning/js/indicators.js
+++ b/djangocms_versioning/static/djangocms_versioning/js/indicators.js
@@ -6,22 +6,18 @@
     function ajax_post(event) {
         event.preventDefault();
         const element = $(this);
-        let csrfToken = '';
-
-        if (element.closest('.cms-pagetree-dropdown-item-disabled').length) {
-            return;
-        }
-        if (window.CMS?.config?.csrf) {
-            // CMS object not always availavle in the admin
-            csrfToken = window.CMS.config.csrf;
-        } else {
+        let csrfToken = window.CMS?.config?.csrf || $('input[name="csrfmiddlewaretoken"]').val();
+        if (!csrfToken) {
+            // Finally try cookies
             const cookieToken = document.cookie.match(/csrftoken=([^;]*);?/);
 
             if (cookieToken && cookieToken.length > 1) {
                 csrfToken = cookieToken[1];
+            } else {
+                showError('CSRF token not found');
+                return;
             }
         }
-        csrfToken = csrfToken || $('input[name="csrfmiddlewaretoken"]').val() || 'no-token';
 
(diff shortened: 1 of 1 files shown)


fsbraun:
@Will-Hoey Could you check if this PR solves your issue?

djangocms_versioning/static/djangocms_versioning/js/indicators.js
```diff
@@ -5,19 +5,28 @@
...
+        let csrfToken = '';
+
         if (element.closest('.cms-pagetree-dropdown-item-disabled').length) {
             return;
         }
-        let csrfToken = document.cookie.match(/csrftoken=([^;]*);?/)[1];
+        if (window.top.CMS.config.csrf) {
+            csrfToken = window.top.CMS.config.csrf;
+        } else {
+            const cookieToken = document.cookie.match(/csrftoken=([^;]*);?/)
```
fsbraun, line 17:
```suggestion
            const cookieToken = document.cookie.match(/csrftoken=([^;]*);?/);
```

fsbraun:
@will-hoey Can you quickly check if this patch solves your issue?

Will-Hoey:
> @Will-Hoey Can you quickly check if this patch solves your issue?

Apologies for the delay. Busy week!

With `CSRF_USE_SESSIONS` as either value - when the admin is in the modal it works as expected, however the same issue occurs when not in the modal, at `admin/news/article/`.  The same console error of `window.top.CMS` is still there.

A side note, and this isn't really likely to be too much of a problem, if I switch that setting and then try any post request (logout, unpublish etc) it'll raise an incorrect token message unless I refresh the page. Kind of expected though!

fsbraun:
@Will-Hoey Ah, right, that makes sense: In the admin (only) panel the CMS object is not available. In that case, the token needs to be pulled from a hidden input field. Do you think you can give it another try?

Will-Hoey:
@fsbraun the new commit works when `CSRF_USE_SESSIONS = False` but not when it's `True`. Same js error as mentioned before

fsbraun:
@Will-Hoey It works for me now. Maybe an outdated token still in the cookies? I've changed the order, so that the token from the form precedes the one in the cookies.

fsbraun:
Thank you so much for the support! Sometimes the small things are more tricky than you expect!

PR #448, opened 2025-02-14
ci: update ruff configuration

### LINKED ISSUE #376: Sync pre-commit checks with django-cms
I've made changes over the weekend to the pre-commit hooks in django-cms.

Ideally we'll have these consistent between projects. The main thing in the change is adopting the ruff linter and formatter.

In theory you should be able to just copy & paste the config from django-cms into these other projects;

https://github.com/django-cms/django-cms/blob/develop-4/.pre-commit-config.yml

The way ruff works is dictated by the `pyproject.toml` config which you can see here;

https://github.com/django-cms/django-cms/blob/develop-4/pyproject.toml

This project has one already, so this isn't as much a copy & paste job. The main thing is to have the same things in the `select` block.

Then in django-cms you'll see it uses `extend-ignore` and `extend-exclude`. So where projects like this on have `exclude` and `ignore`, we want to update these to the `extend-` versions.

Then you'll see a `# TODO fix separately` block. Create one of these that includes all the checks that currently fail. These can be fixed in smaller PRs.

To make sure you can run ruff, it's a python package. So in your virtual env run `python -m pip install ruff`

Also make sure you have pre-commit setup;

```bash
python -m pip install pre-commit
pre-commit install
```

Now any commits you make will get checked by the project's hooks. And if you add more, you just run the `pre-commit install` again.

Now you can run ruff using `ruff check .`

This will run the defined checks against the current directory. It's at this point you need to make judgements on if it's running on any directories that you could exclude because it's not part of the project. And then start adding to that `# TODO` list of rules.

Hopefully that's enough info to get started, but let me know if you've any questions. And of course, you can query things on slack as well. Maybe using the contributors channel for things realted to these changes.

earthcomfy:
## Description

Update ruff configuration to be in sync with `django-cms`. The line length that was already defined here is 120. Let me know if we need to update that to 119. Closes #376 

## Summary by Sourcery

CI:
- Update Ruff configuration to extend existing excludes and ignores instead of overwriting them, and add a TODO to fix remaining issues separately

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [tool.ruff]
-exclude = [
+extend-exclude = [
   ".eggs",
   ".git",
   ".mypy_cache",
@@ -13,7 +13,6 @@ exclude = [
 line-length = 120
 
 [tool.ruff.lint]
-# https://beta.ruff.rs/docs/configuration/
 select = [
   "E",  # pycodestyle errors
   "W",  # pycodestyle warnings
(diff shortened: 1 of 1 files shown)


marksweb (approved):
Thank you 👍

PR #539, opened 2026-04-24
Add Versioning Buttons to Grouper Admin.

Will-Hoey:
## Description
Adds versioning object-tools buttons (Publish / New Draft / Revert /  to grouper admin change forms, mirroring the content.                         Previously the grouper change view at /admin/<app>/<grouper>/<pk>/change/  had no versioning 

As a prerequisite, edit_redirect_view is brought in line with every action view: it now reads ?next= via _internal_redirect (same publish/revert/archive/discard/unpublish use) instead of the query string as opaque forwarding. This lets the new buttons redirect back to the grouper page after creating a instead of jumping to the content admin.

## Related resources

- https://github.com/django-cms/djangocms-versioning/issues/538#issuecomment-4305102089

## Checklist

## Summary by Sourcery

Expose versioning actions in grouper admin change forms and align the edit redirect behaviour with other versioning admin actions.

New Features:
- Add versioning object-tool buttons (Publish, New Draft, Revert, Versions) to grouper admin change forms via a custom change form template and context.
- Introduce a grouper-specific change form template that loads versioning object-tools scripts and styles and injects the buttons into the object-tools area.

Bug Fixes:
- Ensure edit_redirect_view honours a resolvable ?next= parameter consistently with other action views while still creating drafts and avoiding duplicates.
- Fall back to the default editable redirect when ?next= does not resolve to a known admin view, avoiding incorrect external redirects.

Enhancements:
- Expose the resolved content object on grouper admin change views to support rendering versioning controls.
- Refine edit_redirect_view query parameter handling so unrelated GET params are preserved only where appropriate.

Tests:
- Add tests covering edit_redirect_view precedence and fallback behaviour for ?next=, including existing drafts.
- Add tests ensuring grouper change forms render the New Draft control for published content only and expose the versioning content object in the template context.

### DIFF
--- a/djangocms_versioning/templates/djangocms_versioning/admin/grouper/change_form.html
+++ b/djangocms_versioning/templates/djangocms_versioning/admin/grouper/change_form.html
@@ -1,0 +1,17 @@
+{% extends versioning_fallback_change_form_template|default:"admin/change_form.html" %}
+{% load static %}
+
+{% block extrahead %}
+    {{ block.super }}
+    <script src="{% static "djangocms_versioning/js/object-tools.js" %}"></script>
+{% endblock %}
+
+{% block extrastyle %}
+    {{ block.super }}
+    <link rel="stylesheet" href="{% static "djangocms_versioning/css/object-tools.css" %}">
+{% endblock %}
+
+{% block object-tools-items %}
+    {% include "admin/djangocms_versioning/grouper_versioning_buttons.html" %}
+    {{ block.super }}
+{% endblock %}
--- a/djangocms_versioning/admin.py
+++ b/djangocms_versioning/admin.py
@@ -309,6 +309,21 @@ class ExtendedGrouperVersionAdminMixin(ExtendedListDisplayMixin):
 
     """
 
+    change_form_template = "djangocms_versioning/admin/grouper/change_form.html"
+
+    def render_change_form(self, request, context, add=False, change=False, form_url="", obj=None):
+        """Expose the current content object to the change form so the
+        versioning object-tools buttons (Publish / New Draft / Revert /
+        Versions) can be rendered for grouper admins.
+        """
+        if "versioning_fallback_change_form_template" not in context:
+            context["versioning_fallback_change_form_template"] = super().change_form_template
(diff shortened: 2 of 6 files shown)


djangocms_versioning/templates/djangocms_versioning/admin/grouper/change_form.html
Will-Hoey, line 1: The alternative here is to update the existing `versioning_buttons.html` to handle either the Grouper or the Content models. Happy to update if that's the better choice but could break existing implementations/overrides

fsbraun: Yes, I would appreciate this. So future changes are consistent for page contents and custom GrouperAdmins.

fsbraun:
@Will-Hoey Great stuff! Thanks for the PR. 

I would appreciate if we could just have one template for the object tools that works for both page contents and  grouper admins. 

Could it be part of the state indicator mixing? (This mixing can be on GrouperAdmins and content admins.)

Also, I second sorcery's concern about the get params. The admin stores search filters there and they should not be lost.

Will-Hoey:
> @Will-Hoey Great stuff! Thanks for the PR.
> 
> I would appreciate if we could just have one template for the object tools that works for both page contents and grouper admins.
> 
> Could it be part of the state indicator mixing? (This mixing can be on GrouperAdmins and content admins.)
> 
> Also, I second sorcery's concern about the get params. The admin stores search filters there and they should not be lost.

@fsbraun Moving to the StateIndicatorMixin would mean moving/adding a whole `render_change_form` method rather than extending the existing one. Is that okay for the scope of this?

Will-Hoey:
@fsbraun Just checking the scope change before I start the switchover to the other mixin

fsbraun:
@Will-Hoey I was testing versioning against Django 6.1 when I realized Object Tools have been refactored. With the 6.1 fix, I also added versioning object tools for groupers. Can you give it a try run?

New PR: #543

fsbraun:
Implemented in #543

